PR #19, opened 2016-06-22
Fix typo in MarkdownExtensions.cs

jorrit:


xoofx:
Thanks. That's the only typo? :sweat_smile:

jorrit:
I guess not then :) But at least now there is one less.

PR #24, opened 2016-06-23
Update HtmlAttributes CopyTo logic

billboga:
New behavior does an additional check on the `Classes` and `Properties` properties to make sure null-values are not passed to their respective List<T> constructors.
Previous behavior would throw an exception is the former was true and the `share`-parameter was `false`.

Ref. https://github.com/lunet-io/markdig/issues/23.


billboga:
This PR does include a couple tests, but I was not sure how to add them to the existing tests project. So, I just recreated the referenced project's directory structure for the affected file.

xoofx:
Oups very sorry I pushed similar changes at the same time! (Let me know if you start a PR, otherwise I might jump to quickly on it! :sweat_smile: )

Issue #17, opened 2016-06-21
Converting to Markdown


giacomocariello:
I'm trying to develop an alternative renderer which should produce exactly the same content as the file input. I'm wondering what's the easiest way to represent inter-block whitespace. If every block kept slicing info (start, offset), it would be easier to derive whitespace info.

Could you give me some hint on what could be the best approach?

xoofx:
I'm glad that you would like to add this, as I wanted to do it but didn't have enough time!

The general idea I had was to:

- Add a `Markdown.Normalize()` method
- Add a `ToMarkdownRenderer`
- Implement an `ToMarkdownObjectRenderer` for each block/inlines
- Add these renderers to all extensions

Regarding the inter-block, you cannot rely on the info (start, offset), because the syntax tree could be built by code without a parser. You basically need to have a few rendering options (per block type..etc. it will depend. Typically options like cutting paragraph to a certain width of characters...etc.) and a default behaviour for the spaces we should be CommonMark compliant depending on the block types, so that we can reparse the document and get exactly the same HTML rendering.

So a PR for this would be much welcome! You can start by the core parsers, and progressively we could extend it to all supported extensions.

Let me know if you are interested!

giacomocariello:
Hi Alexandre,
my plan was to add a MarkdownRenderer to allow full or partial normalization, where "partial" would mean that, depending on the renderer configuration, some rendering options may be enforced or may only be considered if no parser info is available. I suppose accessing stringslice info for all the blocks, where available, would allow me to offer this kind of flexibility.

WDYT?

At the moment I've already started writing a simple POC which subclasses TextRendererBase and reimplements core block and inline renderers, just to make some practice.

xoofx:
I would prefer first to focus on a plain normalize that doesn't use the original source document but only the syntax tree.
Let me know your progress, you can open a PR on a fork and I can comment as soon as you feel it starts to be reviewable (even for early direction)

madskristensen:
Can the normalizer take options that lets the caller specify indent size (2 spaces instead of 4 for instance)?

giacomocariello:
Sure.

xoofx:
@giacomocariello do you have a fork with already some code? as I would like to implement this feature asap for MarkdownEditor, otherwise I may start it directly in a new markdig branch, let me know!

xoofx:
Closing this issue as it is now tracked by #155

PR #32, opened 2016-06-28
Add support for normalizing Markdown

xoofx:
Related to issue #17 

This is still in wip.
- [x] support all CommonMark core components
  - wip
- [ ] Add normalize options (limit text width....etc.)
- [ ] support all extensions


eamodio:
Is there a blocking issue on PR? Or is there anything we can help out with? This would be a great feature to have -- as I would love to be able to programmatically create a `MarkdownDocument` and be able to generate the markdown string from it.

xoofx:
Nothing blocking, just my limited spare time and lower personal interest for this specific feature.

eamodio:
I'd like too help out on it - could you provide any overview of what is left?

xoofx:
@eamodio sure, you are welcome. The first round is to support the core CommonMark syntax. The current commit changes on my branch contains some work for this (but still some syntaxes are not supported/well supported, like escape characters - I can explain later how to implement it)
Then, once the core part is done, we can start to select candidate for extensions (prioritize first the extensions that are part of the Github Flavored Markdown like pipetables, strikethrough... etc.).
Let me know if you need more details.

tthiery:
I just stumbled over this. I need something similar (I want to analyze the AST, fetch some stuff from it, and then serialize some parts of the AST (normalized or not).

What is the progress? How heavy would this be to complete it? 2 hours, 10 hours, 20 hours?

xoofx:
> What is the progress? How heavy would this be to complete it? 2 hours, 10 hours, 20 hours? Willing to help out.

I have no spare-time to work on this (as I have no personal requirement for this feature), so no progress has been done since I started to prototype the idea. For the core Markdown features (CommonMark only), I guess it requires a few hours of work (with tests...etc.). For all the extensions, quite more.

xoofx:
Feel free to branch from my original PR

tthiery:
I forked the repo and the normalize branch and saw into what is missing. This is my view (correct me if I am wrong)

- rebase to the HEAD of master works
- Renderer for normalization exist in similar scale than for the HtmlRenderer
- Most of the unit test cases are missing.
- There are no test cases for normalization. But maybe that is already covered in the parser tests.
- TODOs for 
  - [ ] List
  - [ ] CodeInline
  - [ ] LineBreakInline
- No options like
  - [ ] Limit Text Width

**Questions** @xoofx

- Are the renderer - ignoring TODOs - covering the complete syntax?
- Extensions are not covered. Maybe better as separate PR?
- Do we need unit tests for the "normalization" test cases? Or is that covered as part of the parser unit tests?
- Is there any list of "normalized" markdown? I checked the CommonMark spec for (as a sample) <http://spec.commonmark.org/0.27/#thematic-breaks>but they do not specify a preferred sample.
- What about LinkReferenceDefinition?

xoofx:
About normalization, it is just options to modify the way is printed back to markdown (like limit the column of paragraphs). You don't have to worry too much of it at this point.

The original PR is very far from complete even for the renderers that were implemented. It was just a bootstrap to check how much work it would require and if there was any changes required to some existing classes (like handling the indent). Most likely there are missing TODO as well.

> Are the renderer - ignoring TODOs - covering the complete syntax?

Can't remember the state of each, so you have to check them. The CommonMark part is quite limited, so it is only a few classes.

> Extensions are not covered. Maybe better as separate PR?

yes

> Do we need unit tests for the "normalization" test cases? Or is that covered as part of the parser unit tests?

we will (for the part specific with options activated, like limit column). One possible automated part for all tests is to output some markdown, re-feed it to the parser, re-output it and check that it doesn't shift.

> Is there any list of "normalized" markdown? 

I don't know, but "normalized" is nothing more than just a few rendering options. We have the freedom to add whatever is relevant for common use cases. We will see which one are valuable. Obviously, limit the width could have an impact on many renderers, so some normalization are more generally intrusive than others (e.g a normalization could work on list numbering for example).

> What about LinkReferenceDefinition?

They should be part of the rendering as well... not sure exactly what do you imply?

> I found some bugs, check this commit file content

Sure, again, the original PR was a very small proof of concept and in the early stage of a WIP.
There will be may be some challenges with detecting characters that need to be escaped also (for the inline part)

tthiery:
I have now unit tests for the straight forward things and fixed the bugs and TODOs so far.

I will go ahead by ...
- cleaning the test suite
- spec the normalization cases
- not implementing the line breaking thing. Too complex for me right now ;) That will also go to an extra PR.

tthiery:
@xoofx can you take a look at https://github.com/tthiery/markdig/blob/normalize/src/Markdig.Tests/TestNormalize.cs

I am curious about your opinion in testing AST to nomalized Markdown? I find the AST very hard to work with (I know.. damn youth...  we enjoyed XLinq to much ;)). Is there a better way to annotate that? Or should we better base testing on alernative-to-normalizet text cases.

xoofx:
> I am curious about your opinion in testing AST to nomalized Markdown? I find the AST very hard to work with (I know.. damn youth... we enjoyed XLinq to much ;)). Is there a better way to annotate that? Or should we better base testing on alernative-to-normalizet text cases.

Yeah, it will be too laborious. I would prefer if we could test as I started by checking plain markdown text:

- Input and renormalize ouput are equal, so just an `AssertNormalize` with only one input (that will be the output used to check the renormalize)
- Input and renormalize output are different: same `AssertNormalize` with the two args (input and output)

tthiery:
Okay, understood. How far should normalization go? I think I will start to support the variations offered by the AST (e.g. different fencing characters for a `FencedCodeBlock`). Later PR should focus on limiting this by (default) settings.

xoofx:
Closing this issue as the first part (core CommonMark) is handled by the PR #154 which was a continuation of this PR.

Will open a separate issue to track the remaining work

PR #36, opened 2016-07-09
Adding support for rel=nofollow in links

synhershko:
This PR adds support for automatically adding rel=nofollow to HTML links - this in many times is desired (if not necessary) especially when allowing content that is not necessarily trusted to be published.

### DIFF
--- a/src/Markdig/Extensions/NoRefLinks/NoRefLinksExtension.cs
+++ b/src/Markdig/Extensions/NoRefLinks/NoRefLinksExtension.cs
@@ -1,0 +1,32 @@
+﻿// Copyright (c) Alexandre Mutel. All rights reserved.
+// This file is licensed under the BSD-Clause 2 license. 
+// See the license.txt file in the project root for more information.
+
+using Markdig.Parsers.Inlines;
+using Markdig.Renderers;
+using Markdig.Renderers.Html.Inlines;
+
+namespace Markdig.Extensions.NoRefLinks
+{
+    public class NoRefLinksExtension : IMarkdownExtension
+    {
+        public void Setup(MarkdownPipelineBuilder pipeline)
+        {
+        }
+
+        public void Setup(IMarkdownRenderer renderer)
+        {
+            var linkRenderer = renderer.ObjectRenderers.Find<LinkInlineRenderer>();
+            if (linkRenderer != null)
+            {
+                linkRenderer.AutoRelNoFollow = true;
+            }
+
+            var autolinkRenderer = renderer.ObjectRenderers.Find<AutolinkInlineRenderer>();
+            if (autolinkRenderer != null)
+            {
+                autolinkRenderer.AutoRelNoFollow = true;
+            }
+        }
+    }
+}
--- a/src/Markdig/Extensions/NoRefLinks/NoRefLinksExtension.cs
+++ b/src/Markdig/Extensions/NoRefLinks/NoRefLinksExtension.cs
@@ -2,7 +2,6 @@
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
 
-using Markdig.Parsers.Inlines;
 using Markdig.Renderers;
 using Markdig.Renderers.Html.Inlines;
 
--- a/src/Markdig/MarkdownExtensions.cs
+++ b/src/Markdig/MarkdownExtensions.cs
@@ -20,6 +20,7 @@ using Markdig.Extensions.Hardlines;
 using Markdig.Extensions.ListExtras;
 using Markdig.Extensions.Mathematics;
 using Markdig.Extensions.MediaLinks;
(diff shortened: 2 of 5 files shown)


synhershko:
FWIW, I did contemplate adding this on the parsers layer, but (1) this seems too HTML specific and (2) change to the renderes would still be required.

src/Markdig/Renderers/Html/Inlines/LinkInlineRenderer.cs
```diff
@@ -52,6 +57,11 @@ protected override void Write(HtmlRenderer renderer, LinkInline link)
             }
             else
             {
+                if (AutoRelNoFollow)
```
xoofx: This should go inside a `if (renderer.EnableHtmlForInline)`

xoofx:
Thanks!

Though, I'm wondering if this extension is really "usable" :sweat_smile: : as It applies this change to all links into a markdown document, even inter-document links within a website would be tagged as nofollow right? Couldn't it be annoying?

Why not using generic attributes on dedicated links that you really don't want to follow like this?

```
This is a [nofollow link](http://boo){rel="nofollow"}
```

synhershko:
Think user content scenarios. I'm moving to using markdig in my blog, and while I want to support markdown in user comments and also auto-link URLs, I prefer they would be nofollow ones - and I don't have control of the markdown they will type.

src/Markdig/Renderers/Html/Inlines/LinkInlineRenderer.cs
synhershko: Gotcha, adding now. I might have missed something - but where is the closing `/>` printed for the `renderer.EnableHtmlForInline` case?

xoofx:
> Think user content scenarios. I'm moving to using markdig in my blog, and while I want to support markdown in user comments and also auto-link URLs, I prefer they would be nofollow ones - and I don't have control of the markdown they will type.

I see, it makes sense now, thanks.

src/Markdig/Extensions/NoRefLinks/NoRefLinksExtension.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Copyright (c) Alexandre Mutel. All rights reserved.
+// This file is licensed under the BSD-Clause 2 license. 
+// See the license.txt file in the project root for more information.
+
+using Markdig.Renderers;
+using Markdig.Renderers.Html.Inlines;
+
+namespace Markdig.Extensions.NoRefLinks
+{
+    public class NoRefLinksExtension : IMarkdownExtension
```
xoofx: Could you add a comment to this?

Also could you add the associated `UseNotRefLinks()` method extension to [MarkdownExtension.cs](https://github.com/lunet-io/markdig/blob/master/src/Markdig/MarkdownExtensions.cs) (don't forget also the `Configure` method that should accept a string like `noreflinks`)

Sorry to bump into these details, once I will have written the doc, that should be easier. 

Thanks again, this is the first community extension! :+1:

xoofx:
Also could you add the associated `UseNotRefLinks()` method extension to [MarkdownExtension.cs](https://github.com/lunet-io/markdig/blob/master/src/Markdig/MarkdownExtensions.cs) (don't forget also the `Configure` method that should accept a string like `noreflinks`)

Sorry to bump into these details, once I will have written the doc, that should be easier. 

Thanks again, this is the first community extension! :+1:

synhershko:
Done. A bit too much ceremony instead of just using MEF or calling by convention - my 2c :)

xoofx:
> Done. A bit too much ceremony instead of just using MEF or calling by convention - my 2c :)

For MEF, I don't want to rely on an external library. Plus I dislike a lot MEF as it is inefficient at scanning assemblies. I'm also not sure about what would be the syntax... but usually, when you have a simple `XXXBuilder` design pattern you don't want to overdesign it with a composition framework...

For the calling convention, you mean by using generic? I thought about this, but I'm also annoyed by the discrepancies that it would introduce. If the extension doesn't have any parameters, a generic `Use<TExt>` could be used, but whenever you have parameters, you would have to introduce specific methods and it would make the API less consistent... 

While it is a bit cumbersome for the library writer, I believe it is a bit more friendly for a end-user...

src/Markdig/MarkdownExtensions.cs
```diff
@@ -59,6 +60,7 @@ public static MarkdownPipelineBuilder UseAdvancedExtensions(this MarkdownPipelin
                 .UseListExtras()
                 .UseTaskLists()
                 .UseDiagrams()
+                .UseNoRefLinks()
```
xoofx: It should not be added to the Advanced extension, as this one is supposed to be somewhat CommonMark compliant and should not change the default behaviour of links

xoofx:
Though one grip with the current design is that `MarkdownExtension.Configure` is not extensible outside of markdig.

PR #43, opened 2016-07-28
Fixes issue where sentences containing strong emphasis words did not correctly delimit the strong tag

christophano:
Scenario:
```
***Strong emphasis*** normal
```
Should result in:
```
<p><strong><em>Strong emphasis</em></strong> normal</p>
```
But actually outputs:
```
<p><strong><em>Strong emphasis</em> normal</strong></p>
```

**Note:** I haven't included a spec to demonstrate the fix, as the specs for this part comes from CommonMark. Would you prefer me to add a new spec to this pull request, or to submit a pull request to the CommonMark spec?


xoofx:
Thanks a lot for this. I have added a test just to track this down

PR #45, opened 2016-08-12
Adds support for rowspans in grid tables

christophano:
I'm afraid I changed the `Grid Table Parser` quite a bit here.
Feel free to wrap my knuckles and suggest better changes though!

This adds support for cells spanning rows, in addition to columns.


christophano:
I'm going to close this temporarily, while I look into more issues relating to the last commit.

xoofx:
Thanks for this, that's nice having rowspan as well!

PR #47, opened 2016-08-19
Fixes issue where trailing text after an inline math block is both included in the math block and appended after the math block.

christophano:
I found an issue where inline math with trailing text both included the text _inside_ the maths block and _outside_ the maths block.

For example:
```
This is a $math block$ with text on both sides.
````

Should render as
```html
<p>This is a <span class="math">math block</span> with text on both sides.</p>
```

But instead we get
```html
<p>This is a <span class="math">math block$ with text on both sides.</span> with text on both sides.</p>
```


xoofx:
Thanks for this fix!

PR #49, opened 2016-08-22
Fixes issue where multiple definition lists are created when they are nested in a list item.

christophano:
I've experienced an issue where a nested definition list outputs multiple definition lists, instead of one list with multiple terms.
The following markdown:
```md
1.  Item One
    
2.  Item Two
    
    Term One
    :   Description One
    
    Term Two
    :   Description Two
    
    Term Three
    :   Description Three
    
3.  Item Three
```
is rendered as:
```html
<ol>
<li><p>Item One</p></li>
<li><p>Item Two</p>
<dl>
<dt>Term One</dt>
<dd>Description One</dd>
</dl>
<dl>
<dt>Term Two</dt>
<dd>Description Two</dd>
</dl>
<dl>
<dt>Term Three</dt>
<dd>Description Three</dd>
</dl>
</li>
<li><p>Item Three</p></li>
</ol>
```
instead of:
```html
<ol>
<li><p>Item One</p></li>
<li><p>Item Two</p>
<dl>
<dt>Term One</dt>
<dd>Description One</dd>
<dt>Term Two</dt>
<dd>Description Two</dd>
<dt>Term Three</dt>
<dd>Description Three</dd>
</dl>
</li>
<li><p>Item Three</p></li>
</ol>
```

I've tracked this down to the way the `DefinitionListParser` will look for the item _before_ the last `ParagraphBlock` for an existing `DefinitionList` but, when the `DefinitionList` is nested in a `ListItem`, there will be an additional `BlankItemBlock` between them.

### DIFF
--- a/src/Markdig/Extensions/DefinitionLists/DefinitionListParser.cs
+++ b/src/Markdig/Extensions/DefinitionLists/DefinitionListParser.cs
@@ -51,8 +51,7 @@ namespace Markdig.Extensions.DefinitionLists
             }
 
             var previousParent = paragraphBlock.Parent;
-            var indexOfParagraph = previousParent.IndexOf(paragraphBlock);
-            var currentDefinitionList = indexOfParagraph - 1 >= 0 ? previousParent[indexOfParagraph - 1] as DefinitionList : null;
+            var currentDefinitionList = GetCurrentDefinitionList(paragraphBlock, previousParent);
 
             processor.Discard(paragraphBlock);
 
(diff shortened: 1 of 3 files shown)


src/Markdig/Extensions/DefinitionLists/DefinitionListParser.cs
```diff
@@ -103,6 +102,19 @@ public override BlockState TryOpen(BlockProcessor processor)
             return BlockState.Continue;
         }
 
+        private static DefinitionList GetCurrentDefinitionList(ParagraphBlock paragraphBlock, ContainerBlock previousParent)
+        {
+            var index = previousParent.IndexOf(paragraphBlock) - 1;
+            if (index < 0) return null;
+            var lastBlock = previousParent[index];
+            if (lastBlock is BlankLineBlock)
```
xoofx: Could we have multiple consecutive BlankLineBlock?

christophano: I could use a while loop to traverse back until it finds anything that _isn't_ a `BlankLineBlock` and returns that as `DefinitionList`, but how would the parent container handle these extra lines? Should they be removed, or could the container handle them?

```c#
private static DefinitionList GetCurrentDefinitionList(ParagraphBlock paragraphBlock, ContainerBlock previousParent)
{
    var index = previousParent.IndexOf(paragraphBlock) - 1;
    if (index < 0) return null;
    var lastBlock = previousParent[index];
    while (lastBlock is BlankLineBlock && index > 0)
    {
        lastBlock = previousParent[--index];
    }
    return lastBlock as DefinitionList;
}
```

xoofx: Don't remember exactly if we can have multiple Blanklines, better to double check.

PR #50, opened 2016-08-23
Changes List Extra to output valid integer for start attribute, instead of roman or latin character.

christophano:
The List Extra extension currently outputs whatever the next value is when continuing lists, which may not be a valid integer.
I've changed it so it converts latin and roman characters to the appropriate number for the start attribute.

### DIFF
--- a/src/Markdig/Helpers/CharHelper.cs
+++ b/src/Markdig/Helpers/CharHelper.cs
@@ -7,6 +7,8 @@ using System.Runtime.CompilerServices;
 
 namespace Markdig.Helpers
 {
+    using System.Collections.Generic;
+
     /// <summary>
     /// Helper class for handling characters.
     /// </summary>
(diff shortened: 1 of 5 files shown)


src/Markdig/Helpers/CharHelper.cs
```diff
@@ -81,6 +86,26 @@ public static bool IsRomanLetterUpperPartial(char c)
         }
 
         [MethodImpl(MethodImplOptionPortable.AggressiveInlining)]
+        public static int RomanToArabic(string text)
```
christophano: Would this be better if we just converted the entire string `ToUpper` at the start?

src/Markdig/Renderers/Html/ListRenderer.cs
```diff
@@ -23,7 +23,7 @@ protected override void Write(HtmlRenderer renderer, ListBlock listBlock)
                     renderer.Write(" type=\"").Write(listBlock.BulletType).Write("\"");
                 }
 
-                if (listBlock.OrderedStart != null && (listBlock.DefaultOrderedStart != listBlock.OrderedStart))
```
christophano: Is `DefaultOrderedStart` needed for anything else?

xoofx:
Good catch, "start" seems to be a number regardless of the list type

Issue #48, opened 2016-08-22
Consider passing mergeIdAndProperties as true in GenericAttributesParser


christophano:
Line 67 of `GenericAttributesParser.cs` calls the `CopyTo` method of the `HtmlAttributes` class.
It passes false as the `MergeIdAndProperties` argument, which means the existing `Id` is overwritten, _even if there is no id on the instance itself_.
If the argument is changed to true, then the `Id` property will only be overwritten if it has been set on the instance.

The scenario that I'm facing right now, is that I have an extension to automatically create Ids for definition term elements (based on their content) but, if I want to also apply a class using the `Special Attributes` extension, the id gets overwritten.

PR #52, opened 2016-08-25
fixes issue where using special attributes overwites element id, even when id is not specified

christophano:
Fixes issue #48


xoofx:
Thanks

PR #55, opened 2016-09-09
Checks if paragraph block is closed when deciding if continuation list is allowable with character other than '1.'

### LINKED ISSUE #54: Problem restarting list after paragraph
The following markdown:
```
1.   First item

Some text

2.   Second item
```
should render as:
```
<ol>
<li>First item</li>
</ol>
<p>Some text</p>
<ol start="2">
<li>Second item</li>
</ol>
```
but instead renders as:
```
<ol>
  <li>
    First item
  </li>
</ol>
<p>
  Some text
</p>
<p>
  2. Second item
</p>
```
This issue is related to the change in CommonMark 0.26 where starting a list after a paragraph is only allowed using `1.` to start to list.
For some reason, Markdig is not respecting the blank line after the paragraph to allow continuation.

christophano:
This fixes #54.
Due to the way `IBlock.IsOpen` is used, it wasn't possible to just check that property, so I have added an internal method to `BlockProcesser` to check if the block is still contained in the `OpenedBlocks` collection.

### DIFF
--- a/src/Markdig.Tests/Specs/ListExtraSpecs.md
+++ b/src/Markdig.Tests/Specs/ListExtraSpecs.md
@@ -105,3 +105,21 @@ iii. Second item
 <li>Second item</li>
 </ol>
 ````````````````````````````````
+
+Lists can be restarted, specifying the start point.
+
+```````````````````````````````` example
+1.   First item
+
+Some text
+
+2.   Second item
+.
+<ol>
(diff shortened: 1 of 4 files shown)


xoofx:
Thanks for this. Might put it as public maybe later. Yeah, IBlock.IsOpen is confusing, only used in some cases, but mostly not up-to-date

PR #56, opened 2016-09-13
Fixes problem with multiple, similar abbreviations not parsing correctly

christophano:
When you have multiple, similar (when an abbreviation is a substring of another abbreviation) abbreviations then only the first abbreviation will be applied.
For example:
```md
*[1A]: First
*[1A1]: Second
*[1A2]: Third
We can abbreviate 1A, 1A1 and 1A2!
```
should result in:
```html
<p>We can abbreviate <abbr title="First">1A</abbr>, <abbr title="Second">1A1</abbr> and <abbr title="Third">1A2</abbr>!</p>
```
but instead we get:
```html
<p>We can abbreviate <abbr title="First">1A</abbr>, 1A1 and 1A2!</p>
```


xoofx:
Thanks, good catch!

PR #57, opened 2016-09-14
Adds exceptions to allow certain punctuation characters to behave as inline delimiters

christophano:
I'm not sure how you'll feel about this one.

The CommonMark spec identifies when whitespace and punctuation can be used as left- and right-flanking delimiter runs. This logic is also used for the extra emphasis extension.
There are cases where certain puntuation characters _should_(?) be allowed.

For example, the minus symbol (or it's more commonly used lookalike, the dash) ought to be allowed as the start of a superscript span:
>
```md
One quintillionth can be expressed as 10^-18^
```
```html
<p>One quintillionth can be expressed as 10<sup>-18</sup></p>
```
><p>One quintillionth can be expressed as 10<sup>-18</sup></p>

Similarly the dagger and double-dagger are commonly used as localised note markers (followed by §, || and #) so should be allowed as the sole contents of a superscript span:
>
```md
Daggers^†^ and double-daggers^‡^ can be used to denote notes.
```
```html
<p>Daggers<sup>†</sup> and double-daggers<sup>‡</sup> can be used to denote notes.</p>
```
><p>Daggers<sup>†</sup> and double-daggers<sup>‡</sup> can be used to denote notes.</p>

**Note 1:** this is implemented as a broad exception for these characters (plus dash, as it is more likely to be used than the actual minus). There is no check for the _type_ of delimiter it is to be used for.

**Note 2:** i have only added the dagger and double dagger to the exceptions. To extend it to include section, parallels and hash would need a small logic change, as parallels uses 2 characters.

### DIFF
--- a/src/Markdig.Tests/Specs/EmphasisExtraSpecs.md
+++ b/src/Markdig.Tests/Specs/EmphasisExtraSpecs.md
@@ -21,6 +21,17 @@ H~2~O is a liquid. 2^10^ is 1024
 .
 <p>H<sub>2</sub>O is a liquid. 2<sup>10</sup> is 1024</p>
 ````````````````````````````````
+ 
+Certain punctuation characters are exempted from the rule forbidding them within inline delimiters
+
+```````````````````````````````` example
+One quintillionth can be expressed as 10^-18^
+
+Daggers^†^ and double-daggers^‡^ can be used to denote notes.
+.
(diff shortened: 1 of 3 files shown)


xoofx:
Good point. We will see how it behaves with this. If there are others characters coming, will report this back to the CommonMark specs and check how to handle this more properly.

PR #58, opened 2016-09-15
Fixes issue when using colon (among others) on lines with abbreviations

christophano:
I discovered a very strange issue when using abbreviations inline after a _potential_ delimiter (in my case, a colon).
The code would throw an `ArgumentException` from line 54 of `ContainerInline.cs`.
Example:
```md
Some abbreviations: HTML, WWW
*[HTML]: Hypertext Markup Language
*[WWW]: World Wide Web
```

This is due to the processor storing the first part of the content in a `LiteralInline` while it tries to validate the character as a valid delimiter. When this fails it returns and continues with the rest of the inline, including the first section. This means that when the `AbbreviationParser` finds a valid abbreviation during the post match, the `LiteralInline` already has a parent, which throws the exception.

The solution involves using the existing `LiteralInline` parent as the container (if it exists) instead of creating a new `ContainerInline` and checking that the parent is null before the call to `ContainerInline.AppendChild`.
There is also a check in `InlineProcessor`, when adding an inline without a parent as a child of the last container that this is not already the last container. This would mean at some point the parent is not being assigned when it ought to have been (I haven't found this yet though), but the check is poss

### DIFF
--- a/src/Markdig/Extensions/Abbreviations/AbbreviationParser.cs
+++ b/src/Markdig/Extensions/Abbreviations/AbbreviationParser.cs
@@ -126,12 +126,13 @@ namespace Markdig.Extensions.Abbreviations
                         // If we don't have a container, create a new one
                         if (container == null)
                         {
-                            container = new ContainerInline()
-                            {
-                                Span = originalLiteral.Span,
-                                Line = originalLiteral.Line,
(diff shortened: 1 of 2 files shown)


xoofx:
Thanks!

PR #60, opened 2016-09-20
Fixes issue where punctuation forms part of word, so abbreviation is not valid

christophano:
I've had an issue where abbreviations are being created for words that contain punctuation.
Example:
```md
*[1A]: First
We should not abbreviate 1.1A
```
Is rendered as:
```html
<p>We should not abbreviate 1.<abbr title="First">1A</abbr></p>
```

I've adjusted the `AbbreviationParser` to allow punctuation to embrace matches only if they are further embraced by \0 or whitespace.


xoofx:
Thanks for digging into these corner cases! Out of curiosity, are you using markdig for a particular project?

christophano:
No problem, this is by far the best markdown parser I've found, and being able to extend it is fantastic.
I am working on a small project that will store content (think multi volume regulations) as markdown with the aim that textual corrections and updates can eventually be done by admin staff rather than development staff.

xoofx:
Thanks for the feedback, that's good to hear. :+1:

PR #68, opened 2016-10-31
Make ContainerBlock's children public

weralabaj:
I need to render anchor links for every header in the page. Since headers are not guaranteed to be unique in the id/name value I include also the name of the parent header. 

For example:
`## heading 1`
`### heading 1.1`

should result in 
```<h3><a name="heading-1-heading-1-1" ...>heading 1.1</a><h3>```

In order to handle that requirement I've created an extension for custom header (`CustomHeadingBlockRenderer : HtmlObjectRenderer<HeadingBlock>`), but I don't see a way to access information about preceding header at a higher level (so in this example for `h3` header I need to access the previous `h2` header).

Seems that if I could access the children of the `ContainerBlock` my issue would be solved:

```
protected override void Write(HtmlRenderer renderer, HeadingBlock obj)
{
    var parentHeader = obj.Parent.children; 
     ...
}
```

Also, if there's a better way to handle that requirement, suggestions will be more than welcome!

### DIFF
--- a/src/Markdig/Syntax/ContainerBlock.cs
+++ b/src/Markdig/Syntax/ContainerBlock.cs
@@ -19,7 +19,7 @@ namespace Markdig.Syntax
     [DebuggerDisplay("{GetType().Name} Count = {Count}")]
     public abstract class ContainerBlock : Block, IList<Block>
     {
-        private Block[] children;
+        public Block[] Children;
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


xoofx:
`ContainerBlock` implements `IList<Block>`  so you really don't need to expose internal Children property as it is already exposed.

weralabaj:
fair enough, thanks!

PR #80, opened 2016-12-15
Adds fluent extension registration method to MarkdownPipelineBuilder.

### LINKED ISSUE #79: Consider introducing generic extension registry method to MarkdownPipelineBuilder
While I like having specific extension methods to load specific extensions onto the pipeline builder, it is also handy not to have to create an extension method for _every_ extension I create.
To that end, I tend to use a single extension method for registering simple extensions.

Would you consider including this method on the `MarkdownPipelinBuilder` class itself? Happy to complete a pull request for this.

Method:
```cs
public MarkdownPipelineBuilder Use<TExtension>() where TExtension : class, IMarkdownExtension, new()
{
    Extensions.AddIfNotAlready<TExtension>();
    return this;
}
```

Usage:
```cs
    .Use<AwesomeExtension>()
    .Build();
```

christophano:
Fixes #79

### DIFF
--- a/src/Markdig/MarkdownPipelineBuilder.cs
+++ b/src/Markdig/MarkdownPipelineBuilder.cs
@@ -91,6 +91,17 @@ namespace Markdig
         internal ProcessDocumentDelegate GetDocumentProcessed => DocumentProcessed;
 
         /// <summary>
+        /// Adds the specified extension to the extensions collection.
+        /// </summary>
+        /// <typeparam name="TExtension">The type of the extension.</typeparam>
+        /// <returns>The current instance of <see cref="MarkdownPipeline" /></returns>
+        public MarkdownPipelineBuilder Use<TExtension>() where TExtension : class, IMarkdownExtension, new()
+        {
+            Extensions.AddIfNotAlready<TExtension>();
+            return this;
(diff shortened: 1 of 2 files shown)


src/Markdig/MarkdownPipelineBuilder.cs
```diff
@@ -91,6 +91,17 @@ public MarkdownPipelineBuilder()
         internal ProcessDocumentDelegate GetDocumentProcessed => DocumentProcessed;
 
         /// <summary>
+        /// Adds the specified extension to the extensions collection.
+        /// </summary>
+        /// <typeparam name="TExtension">The type of the extension.</typeparam>
+        /// <returns>The current instance of <see cref="MarkdownPipeline" /></returns>
+        public MarkdownPipelineBuilder Use<TExtension>() where TExtension : class, IMarkdownExtension, new()
```
xoofx: Couldn't we move this as an extension method to MarkdownExtension to keep it close to other explicit Use?

christophano: Sure.

xoofx:
Thanks for this!

PR #81, opened 2016-12-18
Use Char.ToUpperInvariant instead of Char.ToUpper

volkanceylan:
Due to use of Char.ToUpper, ListExtraItemParser fails in cultures like Turkish where uppercase of 'i' is 'İ'. Char.ToUpperInvariant should be used instead.


xoofx:
Thanks for this fix!

PR #82, opened 2016-12-19
Fixes the IsValidAbbreviation method in AbbreviationParser

christophano:
I made a schoolboy error in my previous fix for the abbreviation parser.
It works perfectly well for _most_ cases, but has issues where an _almost_ valid abbreviation is the first, last or only value in the correct content block.


xoofx:
Could you add a relevant test for this so that we can track it and make sure we don't have regressions?

christophano:
Of course.

xoofx:
Super, thanks!

PR #102, opened 2017-03-19
Lock .NET Core SDK to 1.0.0-preview2

Daniel15:
Without this, Markdig will fail to build on a system that has .NET Core SDK 1.0.0 (final) installed. Currently, Markdig only builds on preview1 and preview2 - You'd need to convert from `project.json` to `csproj` via `dotnet migrate` or Visual Studio 2017 to get it working on the final 1.0.0 SDK.


xoofx:
Thanks! indeed, forgot to do this when reverting VS2017 migration...

PR #103, opened 2017-03-19
Add Replace method to replace an element in an OrderedList

### LINKED ISSUE #101: Utility method for replacing built-in object renderer
Hello! I'm trying to use Markdig to replace MarkdownDeep. Currently I pre-process GitHub-style code fences via a regular expression, as MarkdownDeep does not support them natively. I'd like to remove this as Markdig is nice and supports code fences out-of-the-box, which is awesome. However, I want to output some different HTML to support my current configuration using highlight.js.

Currently I'm doing something like this to replace the built-in CodeBlockRenderer with my own one
```csharp
var originalCodeRenderer = renderer.ObjectRenderers.FindExact<CodeBlockRenderer>();
if (originalCodeRenderer != null)
{
	renderer.ObjectRenderers.InsertAfter<CodeBlockRenderer>(new HighlightCodeBlockRenderer());
	renderer.ObjectRenderers.Remove(originalCodeRenderer);
}
```

This works, but it seems a bit verbose. It would be nice to have a `ReplaceWith` utility method so I could do something like:
```
renderer.ObjectRenderers.ReplaceWith<CodeBlockRenderer>(new HighlightCodeBlockRenderer());
```

I guess I could just have an extension method for this though. Just an idea :)

Thanks for the awesome project!

Daniel15:
Fixes #101

### DIFF
--- a/src/Markdig.Tests/Markdig.Tests.csproj
+++ b/src/Markdig.Tests/Markdig.Tests.csproj
@@ -65,6 +65,7 @@
     <Compile Include="TestHtmlHelper.cs" />
     <Compile Include="TestLineReader.cs" />
     <Compile Include="TestLinkHelper.cs" />
+    <Compile Include="TestOrderedList.cs" />
     <Compile Include="TestPragmaLines.cs" />
     <Compile Include="TestSourcePosition.cs" />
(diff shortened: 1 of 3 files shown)


xoofx:
Thanks!

PR #105, opened 2017-03-29
Treat trailing full stop after a URL as not being part of the URL

MatthewRichards:
A bare URL in a markdown file is interpreted as a URL and rendered accordingly in HTML. However, if the URL is followed by a full stop (.) that full stop is interpreted as being part of the URL. For example consider the URL www.google.com. Because there was a full stop immediately after the URL, Markdig would render this in HTML as `<a href="www.google.com.">www.google.com.</a>`.

This behaviour does not seem terribly unreasonable, given that the trailing dot may constitute a valid URI (see e.g. http://webmasters.stackexchange.com/questions/73934/how-can-urls-have-a-dot-at-the-end-e-g-www-bla-de). However, it's rather inconvenient, and other markdown parsers (such as GitHub's) do not behave this way - see the rendering of the www.google.com link in this comment, above.

I've had a go at implementing this "fix" in Markdig. It's my first look at this source code so I may have done something unpleasant by mistake, but all the tests pass. Basic approach:

* Add a lookahead (`PeekChar`) feature to `ICharIterator`
* If we find a "." while parsing a URL, peek ahead and if that "." would be the end of the URL, terminate parsing now without including the "."

I chose to implement a single-char peek function, partly because that's all we need, partly because it felt most natural in `ICharIterator`, and partly because although `StringSlice` already has an arbitrary-offset peek it looked rather fiddly to implement the equivalent in `StringLineGroup.Iterator`.

Feedback welcomed! I'm happy to have a go at implementing any further improvements necessary, if you agree with the basic premise but have concerns over the implementation.

### DIFF
--- a/src/Markdig.Tests/TestLinkHelper.cs
+++ b/src/Markdig.Tests/TestLinkHelper.cs
@@ -31,6 +31,16 @@ namespace Markdig.Tests
         }
 
         [Test]
+        public void TestUrlTrailingFullStop()
+        {
+            var text = new StringSlice("http://google.com.");
+            string link;
+            Assert.True(LinkHelper.TryParseUrl(ref text, out link));
+            Assert.AreEqual("http://google.com", link);
+            Assert.AreEqual('.', text.CurrentChar);
+        }
+
+        [Test]
         public void TestUrlNestedParenthesis()
         {
             var text = new StringSlice("(toto)tutu(tata) nooo");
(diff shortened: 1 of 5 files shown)


src/Markdig.Tests/TestLinkHelper.cs
```diff
@@ -31,6 +31,16 @@ public void TestUrlUrl()
         }
 
         [Test]
+        public void TestUrlTrailingFullStop()
+        {
+            var text = new StringSlice("http://google.com.");
```
xoofx: Could you add another case with the space character at the end? (can be in the same test method)... even if the code checking for the end is shared with IsEndOfUri...

xoofx:
Thanks for this PR! overall it looks good and changes are quite reasonable. I'm travelling so don't want to merge while I'm not on my working computer, but will do this by the end of the week.

src/Markdig.Tests/TestLinkHelper.cs
MatthewRichards: Good idea, done.

Issue #108, opened 2017-04-28
EmphasisInlineRenderer ignores attributes


aKzenT:
We have a custom parsing module that adds attributes to an EmphasisInline. Unfortunately the renderer does not write any attributes. Is there a specific reason for this? Could this be added?

Thanks!

xoofx:
It is just an oversight ([here](https://github.com/lunet-io/markdig/blob/1bc9d9083c80733f7172997b3d61203d3efe67a1/src/Markdig/Renderers/Html/Inlines/EmphasisInlineRenderer.cs#L40)  is the missing writing attributes). PR Welcome!

aKzenT:
Done!

PR #109, opened 2017-04-28
Fix for issue #108

aKzenT:


xoofx:
Thanks!

aKzenT:
My pleasure. Do you have plans already when this will be released on nuget?

Issue #115, opened 2017-05-24
Static check errors.


Backs:
I checked solution with PVS-Studio and found some problems:

High level errors:
V3095 The 'container' object was used before it was verified against null. Check lines: 154, 178. AbbreviationParser.cs 154
V3095 The 'table.ColumnDefinitions' object was used before it was verified against null. Check lines: 40, 88. HtmlTableRenderer.cs 40
V3095 The 'columnSlice.BlockProcessor' object was used before it was verified against null. Check lines: 198, 203. GridTableParser.cs 198
V3022 Expression '(c < 'A' || c > 'Z') && (c < 'a' && c > 'z')' is always false. HtmlHelper.cs 620
V3022 Expression 'endState == 1' is always true. HeadingBlockParser.cs 98
V3022 Expression 'state > 0 && builder.Length >= 32' is always false. LinkHelper.cs 178
V3022 Expression 'state > 0' is always false. LinkHelper.cs 196
V3022 Expression 'pc == '.' || pc == '-'' is always false. LinkHelper.cs 250
V3022 Expression 'NewBlocks.Count > 0' is always false. BlockProcessor.cs 621

Medium level errors:
V3072 The 'AutoIdentifierExtension' class containing IDisposable members does not itself implement IDisposable. Inspect: headingWriter. AutoIdentifierExtension.cs 20
V3063 A part of conditional expression is always true if it is evaluated: state <= 0. LinkHelper.cs 165
V3062 An object 'openDelimiter' is used as an argument to its own method. Consider checking the first actual argument of the 'MoveChildrenAfter' method. EmphasisInlineParser.cs 331

Low level errors:
V051 Some of the references in 'Markdig.WebApp' project are missing or incorrect. The analysis results could be incomplete. Consider making the project fully compilable and building it before analysis.  0
V3022 Expression 'simpleBench' is always false. Program.cs 100
V3024 An odd precise comparison: tableColumnDefinition.Width != 0.0f. Consider using a comparison with defined precision: Math.Abs(A - B) > Epsilon. HtmlTableRenderer.cs 31
V3024 An odd precise comparison: tableColumnDefinition.Width != 1.0f. Consider using a comparison with defined precision: Math.Abs(A - B) > Epsilon. HtmlTableRenderer.cs 31
V3111 Checking value of 'element' for null will always return false when generic type is instantiated with a value type. OrderedList.cs 27
[...]

PR #116, opened 2017-05-24
Some condition fixes. Issue #115

Backs:
I fixed some obvious errors, that were found by PVS

### DIFF
--- a/src/Markdig/Helpers/HtmlHelper.cs
+++ b/src/Markdig/Helpers/HtmlHelper.cs
@@ -24,7 +24,7 @@ namespace Markdig.Helpers
                 {
                     EscapeUrlsForAscii[i] = $"%{i:X2}";
                 }
-                else if ((char) i == '&')
+                else if ((char)i == '&')
                 {
                     EscapeUrlsForAscii[i] = "&amp;";
                 }
@@ -284,7 +284,7 @@ namespace Markdig.Helpers
                 text.NextChar() == 'D' &&
                 text.NextChar() == 'A' &&
                 text.NextChar() == 'T' &&
-                text.NextChar() == 'A' && 
+                text.NextChar() == 'A' &&
                 (c = text.NextChar()) == '[')
             {
                 builder.Append("CDATA[");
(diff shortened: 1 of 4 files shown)


src/Markdig/Parsers/HeadingBlockParser.cs
```diff
@@ -95,26 +95,23 @@ public override BlockState TryOpen(BlockProcessor processor)
                         }
                         endState = 1;
                     }
-                    if (endState == 1)
```
xoofx: Removing this change completely the meaning of the loop. This is not a valid change.

src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -617,12 +617,6 @@ private void TryContinueBlocks()
                     {
                         leaf.AppendLine(ref Line, Column, LineIndex, CurrentLineStartPosition);
                     }
-
```
xoofx: This is a safe guard that should not be removed. This was put in purpose so that if an extension forgot to add a NewBlock, we throw an error.

src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -247,10 +246,6 @@ public static bool TryParseAutolink(ref StringSlice text, out string link, out b
                     }
                     else if (c == '.')
                     {
-                        if (pc == '.' || pc == '-')
```
xoofx: Don't remember exactly the reasoning (no comments unfortunately), but better fix by adding `pc = c` at the end of the loop rather than removing it all.

src/Markdig/Helpers/HtmlHelper.cs
```diff
@@ -617,7 +617,7 @@ public static int ScanEntity(string s, int pos, int length, out string namedEnti
             {
                 // expect a letter and 1-31 letters or digits
                 c = s[pos + 1];
-                if ((c < 'A' || c > 'Z') && (c < 'a' && c > 'z'))
+                if (!((c >= '0' && c <= '9') || (c >= 'A' && c <= 'Z') || (c >= 'a' && c <= 'z')))
```
xoofx: yes but not adding `(c >= '0' && c <= '9')` (comment is quite troubling: It is a letter, followed by letters or digits

src/Markdig/Parsers/BlockProcessor.cs
xoofx: Will add a comment above for this.

xoofx:
Thanks, but I have pushed slightly different changes.

Backs:
Ok, no problems. I'm sure, you better know what to do!

PR #118, opened 2017-05-27
GenericAttributes: The first character of the second boolean attribute is skipped

meziantou:
The following markdown:
````md
# Heading{prop1 prop2}
````
output:
````html
<h1 prop1 rop2>Heading</h1>
````

The second attribute `prop2` is not correctly added to the output.


xoofx:
Good catch, thanks a lot!

PR #119, opened 2017-05-28
MediaLinkExtension: Remove unexpected closing tag

meziantou:
The following markdown:
````
![Video3](https://sample.com/video.mp4)
````
Ouputs:

````html
<p><video width="500" height="281" controls><source type="video/mp4" src="https://sample.com/video.mp4"></source></video></iframe></p>
````

The `</iframe>` is unexpected as there is no opening tag.


xoofx:
Good catch, thanks!

PR #121, opened 2017-06-14
Fix issue in GenericAttributesParser causing inline elements to be sk…

pearsonn:
…ipped

When attempting to apply classes to a table, I found all the text within the table was missing.  The line I changed was causing the table element to be skipped because the container is modified while MarkdownParser.ProcessInlines() is looping over the container items.


xoofx:
Thanks!

PR #123, opened 2017-06-28
Fix abbreviation at start of paragraphs

joannaz:
There was an issue with an abbreviation term being at the start of the paragraph, in which the index does not update and therefore words can be repeated. This fixes it!


clarkd:
+1 for this!

xoofx:
Thanks!

PR #124, opened 2017-06-29
Fix abbreviation matching

joannaz:
There was an issue with abbreviations matching on a section of a word. For example: PR was matching on PRAA.


xoofx:
Seems some tests are failing...

joannaz:
Sorry! I have fixed them now.

xoofx:
Thanks! btw, could you add the relevant tests to [AbbreviationSpecs.md](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/AbbreviationSpecs.md) (you have to regenerate the tests by saving the `Specs.tt` t4 file) for the problem you had in the first place so that we can catch any regressions/behavior on that?

joannaz:
Alright, just pushed the tests up- thanks a bunch!

PR #139, opened 2017-08-29
Implement Pandoc compatible front matter parsing

### LINKED ISSUE #138: Support for Pandoc YAML front matter
YAML front matter support was added to resolve #37. Wanted to see about adding support for Pandoc-flavored YAML front matter. That flavor allows either leading dots (...) or leading dashes (---). Similarly, it can end with either dots or dashes; it is not required that the end match the start of the fence. The most common pattern with Pandoc seems to be leading with dashes and ending with dots. I'm trying to help improve the support for Pandoc markdown previews in Markdown Monster.

I can provide a PR on this -- either adding it to the main front matter parser or as a separate extension that can be enabled.

kenmuse:
Resolves #138

### DIFF
--- a/src/Markdig/Extensions/Yaml/YamlFrontMatterParser.cs
+++ b/src/Markdig/Extensions/Yaml/YamlFrontMatterParser.cs
@@ -1,44 +1,128 @@
 // Copyright (c) Alexandre Mutel. All rights reserved.
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
+using Markdig.Helpers;
 using Markdig.Parsers;
+using Markdig.Syntax;
 
 namespace Markdig.Extensions.Yaml
 {
     /// <summary>
     /// Block parser for a YAML frontmatter. 
     /// </summary>
-    /// <seealso cref="Markdig.Parsers.FencedBlockParserBase{YamlFrontMatterBlock}" />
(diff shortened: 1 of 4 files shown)


src/Markdig/Extensions/Yaml/YamlFrontMatterParser.cs
```diff
@@ -1,44 +1,128 @@
...
+    public class YamlFrontMatterParser : BlockParser
     {
-        // We reuse a FencedCodeBlock parser to grab a frontmatter, only active if it happens on the first line of the document.
+		// We reuse a FencedCodeBlock parser to grab a frontmatter, only active if it happens on the first line of the document.
 
-        /// <summary>
-        /// Initializes a new instance of the <see cref="FencedCodeBlockParser"/> class.
-        /// </summary>
-        public YamlFrontMatterParser()
+		/// <summary>
```
xoofx: spaces instead of tabs for indentation for the whole file

xoofx:
LGTM

src/Markdig/Extensions/Yaml/YamlFrontMatterParser.cs
kenmuse: Corrected. Thanks. Have you considered adding an [.editorconfig](http://editorconfig.org/) to the root of the project to help make sure IDE's and tools are aware of the preference?

xoofx: sure I will

Issue #104, opened 2017-03-24
Strip markdown


ProsperoTheBun:
Feature Request: 
It would be useful to be able to remove all markdown without converting to HTML. E.g, for a preview of a blog's body text.

So an input like ...
```
__Reasons to be happy__
- Sunshine
- Meerkats
- Football
```

Would produce 
```
Reasons to be happy
Sunshine
Meerkats
Football
```

ProsperoTheBun:
FYI, I workaround this on webpages by using jQuery to read the "text" of the element then write it back in. 

```
var text = $("#body-snippet").text();
$("#body-snippet").text(text);
```

But for my app feed API, I'd rather do this server-side.

leotsarev:
Note that's a good practice to sanitize HTML anyway (as markdown parsers do not sanitize). Most of HTML sanitizers could remove all html anyway

xoofx:
> It would be useful to be able to remove all markdown without converting to HTML. E.g, for a preview of a blog's body text.

Markdown at its first origin is text oriented, so that even a simple bullet point list would be readable in plain text. So in your examples above, removing the `-` for the list even for displaying as text is quite an opinionated "preview"... For other markdown elements, I can also see quite different ways to output a text preview (emphasis could remove, but links? still displaying the link or not?...etc.)

Personally don't have time nor special interest in this feature, but a PR is welcome.

Note that `HtmlRenderer` has already a property `EnableHtmlForInline`, that is currently working only for inlines. If someone wants to work on this PR, you can still use the HtmlRenderer, but add a `EnableHtmlForBlock`, and revisit all html block renderers (like the `HeadingRenderer`) and enclose with if/else the portions that are outputing HTML (but keep at least the `WriteLine` to keep a reasonable structure of the text)

xoofx:
Errata about my previous comment: If someone wants to provide such feature, I would much prefer that we do this work as part of #32, as it covers broader use cases.

xoofx:
Actually, scratch that... #32 is probably not ideal either if we want to keep only text... Let me think about this...

xoofx:
So this should be fixed by PR #140 that was just merged to master

xoofx:
So the workaround for this issue is to use a `HtmlRenderer` and setting the `EnableHtmlForInline` to `false`

leotsarev:
I submitted public API facade for this. I wonder if you are open to change like that, I'm happy to expand if required.

PR #140, opened 2017-09-12
Added EnableHtmlForBlock flag on HtmlRenderer for issue #104

peinearydevelopment:
#104 I would really like to see this feature as well. From my understanding of your comments, these changes are what is necessary to enable it. Please accept this contribution to enable the feature or let me know what else needs to get done to enable it.

Thanks for creating such a great/useful library!!!

### DIFF
--- a/src/Markdig/Renderers/Html/CodeBlockRenderer.cs
+++ b/src/Markdig/Renderers/Html/CodeBlockRenderer.cs
@@ -42,29 +42,50 @@ namespace Markdig.Renderers.Html
 
                 // We are replacing the HTML attribute `language-mylang` by `mylang` only for a div block
                 // NOTE that we are allocating a closure here
-                renderer.Write("<div")
-                    .WriteAttributes(obj.TryGetAttributes(),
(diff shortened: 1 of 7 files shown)


xoofx:
Hi, unfortunately, I can't accept this PR. Changing the behavior of the Html renderer to not render Html at all is not a good direction (note that the `EnableHtmlForInline` was introduced to support a required feature of rendering nodes without decoration inside links)

The reason is that in order to support #104, I would much prefer that we do this work as part of #32, as it covers broader use cases.

peinearydevelopment:
Thanks for the feedback. That is a little disappointing though as this is for a project I'm currently in the middle of. Please let me know if there is something else I can do to help with this!

xoofx:
> That is a little disappointing 

As I said above, making this as part of the html renderer might not be the worse solution after all.  😉  I just need a bit of time to take a look at the details...

xoofx:
Thanks, I have merged these changes to master

peinearydevelopment:
Thank you for all of your hard work on this library. It is greatly appreciated!!!

PR #147, opened 2017-09-28
media extension template and some refactoring

MaddalenaTech:
Allowing to put a template as iframe instead of just choosing heigth and width.
This allow compatibility with frameworks like bootstrap 4.

http://getbootstrap.com/docs/4.0/utilities/embed/


xoofx:
Could you split the real commit 77d6f42 related to this PR, and the others that are unrelated, otherwise I can't merge it. In the other, there are some changes that I don't want to be applied (even for example if a variable is not used)

PR #149, opened 2017-10-17
Nomnoml support

maulwuff:
added support for nomnoml diagrams


xoofx:
Oh, an oversight, but for the next time (as this is not well documented), whenever you change a spec, you need to re-save the file `Specs.tt` in the solution so that tests are actually updated with latest changes

PR #152, opened 2017-10-24
Normalize Bugfixes and extension of Unit Tests

tthiery:
See attached the unit tests and bugfixes for the normalization renderer.


tthiery:
Was too messy. Let me figure out what happened.

PR #154, opened 2017-10-24
Normalize Bugfixes and Tests

tthiery:


src/Markdig/Renderers/Normalize/ListRenderer.cs
```diff
@@ -37,7 +37,7 @@ protected override void Write(NormalizeRenderer renderer, ListBlock listBlock)
                     renderer.Write(index + "");
                     renderer.Write(listBlock.OrderedDelimiter);
                     renderer.Write(' ');
-                    renderer.PushIndent("  ");  // TODO: output accurate indent
+                    renderer.PushIndent(new string(' ', IntLog10Fast(index) + 3));
```
xoofx: Not sure to understand the logic here. Why a higher index would generate an indent?

tthiery: because "10." has less characters than "123456."

xoofx: Ah ok, so you want to left pad with spaces.. isn't something that should be better part of a normalizing stuffs and not the default one? (Not sure I align usually list items by left padding with spaces...)

tthiery: image

xoofx: han ok sorry, a bit slow today 😅

src/Markdig/Renderers/Normalize/ListRenderer.cs
```diff
@@ -62,7 +62,20 @@ protected override void Write(NormalizeRenderer renderer, ListBlock listBlock)
             }
 
             renderer.CompactParagraph = compact;
-            renderer.WriteLine();
+
+            renderer.FinishBlock();
         }
+
+
+        public int IntLog10Fast(int input) =>
```
xoofx: Make this method private static instead

xoofx:
Cool. So to recap, except the escape characters that are missing (I will add this in a separate commit, as I have to slightly change the EscapeInlineParser and Inline node for this), are all core CommonMark nodes all handled by this PR?

tthiery:
To my very humble understanding, yes. I mean you coded most of it, I just fixed the bugs I found and extended the unit tests.

PR #156, opened 2017-10-27
Add simple helper that enables conversion to plain text

### LINKED ISSUE #104: Strip markdown

leotsarev:
Follow up of work on #140 that gives users easy interface to strip markdown without adding any HTML.
Fixes #104 even more

### DIFF
--- a/src/Markdig.Tests/Markdig.Tests.csproj
+++ b/src/Markdig.Tests/Markdig.Tests.csproj
@@ -62,6 +62,7 @@
     <Compile Include="TestLineReader.cs" />
     <Compile Include="TestLinkHelper.cs" />
     <Compile Include="TestOrderedList.cs" />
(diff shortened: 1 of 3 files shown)


xoofx:
Thanks!

leotsarev:
Thanks for you for your all hard work!
Do you have any specific plans about releasing NuGet package with this and #140 ?

xoofx:
> Do you have any specific plans about releasing NuGet package with this and #140 ?

I will try to release a new version later today.

PR #169, opened 2017-11-14
fix autolink extension issue when url ending with comma

yishengjin1413:
Please see the unit test as an example, the original rendering result is `<p><a href="http://your.company.abc,">http://your.company.abc,</a> abc</p>`.

All the test cases are passed with this change.

### DIFF
--- a/src/Markdig.Tests/TestParser.cs
+++ b/src/Markdig.Tests/TestParser.cs
@@ -20,6 +20,12 @@ namespace Markdig.Tests
         }
 
         [Test]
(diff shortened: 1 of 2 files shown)


xoofx:
thanks, though, before merging this, I might actually have a look at https://github.github.com/gfm/#autolinks-extension- as it is the one coming from github that they are using in there cmark fork, so will check if we can use these specs instead and implements this accordingly.

xoofx:
Fixed by commit 964538e

PR #170, opened 2017-11-20
fix yaml frontmatter issue when ending with a empty line

yishengjin1413:
@xoofx Please see the test case. We got another issue after fix [issue160](https://github.com/lunet-io/markdig/issues/160).


xoofx:
Good catch, thanks

yishengjin1413:
@xoofx When do you plan to update the package with this fix?

xoofx:
Just bump the version, should come soon. I don't always push on each PR, so let me know if you are waiting for an update. 😉

yishengjin1413:
Thanks for you so quick response.

PR #173, opened 2017-11-23
Fix bug in LinkHelper.UrilizeAsGfm

qinezh:
The heading id of `# _a_` should be `a` instead of `_a_` in GFM. See https://babelmark.github.io/?text=%23+_a_

### DIFF
--- a/src/Markdig.Tests/TestLinkHelper.cs
+++ b/src/Markdig.Tests/TestLinkHelper.cs
@@ -409,5 +409,13 @@ namespace Markdig.Tests
         {
             Assert.AreEqual(expectedResult, LinkHelper.Urilize(input, false));
         }
+
+        [TestCase(".bc", "bc")]
+        [TestCase("a-.-", "a--")]
+        [TestCase("_a_", "a")]
+        public void TestUrilizeAsGfm_Simple(string input, string expectedResult)
+        {
+            Assert.AreEqual(expectedResult, LinkHelper.UrilizeAsGfm(input));
+        }
     }
 }
--- a/src/Markdig/Helpers/LinkHelper.cs
+++ b/src/Markdig/Helpers/LinkHelper.cs
@@ -104,7 +104,7 @@ namespace Markdig.Helpers
             for (int i = 0; i < headingText.Length; i++)
             {
                 var c = char.ToLowerInvariant(headingText[i]);
-                if (char.IsLetterOrDigit(c) || c == ' ' || c == '-' || c == '_')
+                if (char.IsLetterOrDigit(c) || c == ' ' || c == '-')
                 {
                     headingBuffer.Append(c == ' ' ? '-' : c);
                 }


xoofx:
Hm, that's not what I can see in the current implementation of GFM used on GitHub (not the cmark-gfm). Take a look at this link: https://github.com/lunet-io/scriban/blob/master/doc/builtins.md#arrayadd_range for example, it doesn't remove _

Also the official implementation https://github.com/jch/html-pipeline/blob/master/lib/html/pipeline/toc_filter.rb (used for rendering the page above), is using the regexp, with `\w` which does keep underscores.

For me, it looks more a bug in cmark-gfm. They are unlikely going to break existing links like this all other the github repos...

xoofx:
We can follow this at https://github.com/github/cmark/issues/68

xoofx:
Note that I took some time to add clean support for GFM auto-identifiers when writing the [doc of scriban](https://github.com/lunet-io/scriban/tree/master/doc), as I was auto-generating the links with markdig and wanted to be sure that all the doc links on GitHub were actually working.

qinezh:
Thanks for the information, I agree with you that it may be a bug in cmark-gfm...

xoofx:
So back on the subject, it seems a bug in the auto-identifier. It should actually render the title by discarding actually emphasis before urilizing (thought it did, but maybe not, will double check)

qinezh:
It's ok to close this PR if you find that the bug should be fixed in somewhere else. : )

xoofx:
@qinezh Yep, I will close it as soon as I have the fix pushed.

xoofx:
Ok, so @qinezh it seems that I completely missed the point of your original PR.

The function `LinkHelper.UrilizeAsGfm` is not used as-is by the auto-identifier extension. It expects first the link to have been processed by the markdown emphasis...etc. (so a link is actually rendered with the HTML renderer without the HTML tags):

https://github.com/lunet-io/markdig/blob/fbc8a4116c80873dea2e086610703d37186e2712/src/Markdig/Extensions/AutoIdentifiers/AutoIdentifierExtension.cs#L172-L179

So `LinkHelper.UrilizeAsGfm` was not intended to be used alone. If you check on babelmark, a heading `# _abc_` is correctly generating a `abc` id.

Closing for now as it is the correct behavior, let me know if there is an issue.

xoofx:
Seeing how it is used in MarkdigEngine for docfx:

https://github.com/docascode/markdown/blob/15ffb6836493f08a1e480444c3fae02e84d84d45/MarkdigEngine.Extensions/DFMHeadingId/DFMHeadingIdExtension.cs#L113-L119

I can see why... but that's not the way auto-identifier extension is actually working in markdig. Also, I'm wondering why you are not using the default extension actually?

qinezh:
Since there're some difference between heading id generated by DFM and markdig(advanced), and we think it maybe by design for the implementation of markdig.

For example: 

`# 1. title` will be rendered as:

markdig(advanced): 

```html
<h1 id="title">
  1. title
</h1>
```

dfm:

```html
<h1 id="1-title">
  1. title
</h1>
```

github:

```html
<h1>
  <a id="user-content-1-title" class="anchor" href="#1-title" aria-hidden><span aria-hidden class="octicon octicon-link"></span></a>1. title
</h1>
```

xoofx:
But if you use the option `AutoIdentifierOptions.GitHub` you should get the same behavior than DFM right?

qinezh:
@xoofx Yes, you're right! Thanks for the kindly reminder.

Issue #155, opened 2017-10-24
Add support for normalize - markdown document rountrip


xoofx:
Followup of #17 and #32 and #154 

This is an issue to track the remaining work to handle full markdown document round-trip

- [x] Support all CommonMark core components (#154)
  - [ ] Handle escape characters
- [ ] Add normalize options (limit text width....etc.)
- [ ] Support all extensions

cc: @tthiery

xoofx:
So I have fixed a few cases with list for better handling loose list. 
I have added support for escape characters.
I have also started to add [`NormalizeOptions`](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Renderers/Normalize/NormalizeOptions.cs) to control a bit the markdown output.

The limit text width will require a bit more infrastructure work on the `TextRendererBase`, still nothing done for this (wip)

xoofx:
I have added support for HtmlBlock renderer. The normalize is able to process the [CommonMark specs](https://raw.githubusercontent.com/commonmark/CommonMark/master/spec.txt) and the output is almost identical to the original which is already a good start.

duncanawoods:
`TaskLists` are marked as done but normalise actually strips the task syntax:

``` csharp
    static class Program
    {
        public static MarkdownDocument Parse(string markdown) =>
            Markdown.Parse(
                markdown, 
                new MarkdownPipelineBuilder()
                    .UseTaskLists()
                    .Build());
        
        public static string ToText(this MarkdownObject o)
        {
            using var stringWriter = new StringWriter();
            
            new NormalizeRenderer(stringWriter)
                .Render(o);
            
            return stringWriter.ToString();
        }
        
        static void Main(string[] args) =>
            Console.WriteLine(
                Program.Parse("- [ ] example task")
                    .ToText());
    }
```
prints 

    -  example task

duncanawoods:
Ah, I see, have to use the same pipeline:

``` csharp
    static class Program
    {
        public static MarkdownPipeline Pipeline =>
            new MarkdownPipelineBuilder()
                .UseTaskLists()
                .Build();
        
        public static MarkdownDocument Parse(string markdown) =>
            Markdown.Parse(
                markdown, 
                Program.Pipeline);
        
        public static string ToText(this MarkdownObject o)
        {
            using var stringWriter = new StringWriter();
            
            var renderer = new NormalizeRenderer(stringWriter);
                
            Program.Pipeline.Setup(renderer);
                
            renderer.Render(o);
            
[...]

PR #186, opened 2017-12-17
Add Normalization Support for AutoLinks

tthiery:
Part of #155 

Question here: Are we normalizing auto links to the InlineLink representation OR back to their original form.

If we want to go back to the original form, we need to extend the syntax (IMHO)

I am for normalizing to InlineLinks.

### DIFF
--- a/src/Markdig.Tests/TestNormalize.cs
+++ b/src/Markdig.Tests/TestNormalize.cs
@@ -420,6 +420,15 @@ This is a last line";
             AssertNormalizeNoTrim("**Hello World AB-1**");
         }
 
+        [Test]
+        public void AutoLinks()
+        {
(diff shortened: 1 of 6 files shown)


xoofx:
> Question here: Are we normalizing auto links to the InlineLink representation OR back to their original form.
> If we want to go back to the original form, we need to extend the syntax (IMHO)

We should support both (via options). if the link doesn't contain this information, we have to add it so that we can recover from.

tthiery:
I added a `InlineLink.IsAutoLink` flag to the syntax and added a `NormalizeOptions.ExpandAutoLinks` option.

The `NormalizeAutoLinkRenderer` has a customized Accept method checking these two flags and execute if applicable.

Set the default for the option to true (meaning it will expand to the "default" InlineLink representation).

Ready for Review

xoofx:
Thanks, will give a look next week!

xoofx:
Thanks

PR #189, opened 2018-01-01
Media link extension to support MP3 by default

markheath:
I presume MP3 wasn't intentionally left out this list?


xoofx:
Good catch, thanks!

PR #191, opened 2018-01-08
Improve discoverability of test in VS test runner

leotsarev:
I'm using VS2017 test runner to run unit tests from this solution, and it has names like "Example0??". It's really confusing.

I fixed names to explicitly include name of spec file, so it starts looking like this (see s/shot)
default


xoofx:
Will have to check as I'm using R# and the names may appear differently there...

leotsarev:
It will be like this
default

Note that I could not get my R# to run markdig unit tests (and still have to figure why...)

PR #194, opened 2018-01-08
[WIP] Make tt generator automatically run on build and exclude Specs.cs from source control

### LINKED ISSUE #192: Find a way to regenerate Spec.cs when Specs\*.md changes
1. Change some spec (like Specs\MediaSpecs.md)
2. Rebuild project

Expected result: Specs.cs was automatically updated
Actual result: You have to manually run Specs.tt custom tool to perform that.

P.S. Also Specs.cs probably should be excluded from source control

leotsarev:
fixes #192


xoofx:
I'm unable to get this working with R#, getting just a bunch of "inconclusive: Test not run", while the vs tests are working fine...

xoofx:
Also forgot one issue with this: Specs.tt download the CommonMark specs from a HTTP, but you can't run tests if you are not connected. quite a deal breaker. Now that I recall why I kept the tt files and cs generated was actually for this reason.

leotsarev:
Hm.

leotsarev:
Idea is:
 - separate .tt file which purpose is to download CommonMark Spec
 - if CommonMark spec is already downloaded, this is basically no-op
 - you have to have internet connection once to download CommonMark spec, but you already have to have because of NuGet.

I'm not sure that's optimal solution. I don't like that we rely on specific order for this .tt to execute

leotsarev:
and yes, this is failing on Appveyor. I'll see if I have more ideas, plz leave everything open :-)

xoofx:
Hm, not sure I'm going to accept this PR in the end. Imo, it brings more trouble than what it wants to fix (and as I have never been annoyed by the manual process of re-saving the tt files). It brings several point of failures and actually more bugs, that would require more checks (like downloading a copy of the tt would require to check that once downloaded, if we update the link, it should download it again....etc.

Gonna try another approach to mitigate your original problem by ensuring in the test that if a md files is modified and the tt files was not saved, you get at least an error.

leotsarev:
Resaving .tt files is fine, re-merging Specs.cs is not.
But yes, I'm understand and share your doubts about this approach.

PR #195, opened 2018-01-09
AutoLinkParser should to remove mailto: in outputted text

### LINKED ISSUE #177: AutoLinkParser option to remove mailto: in outputted text
With the AutoLinkParser the outputted text is fine for everything related to an url, this is readable by an user. But, for the an email address, it's strange there is a mailto: prefix in the visible text as well.
Could this be removed? Or optional?

leotsarev:
fixes #177


xoofx:
thanks!

Issue #143, opened 2017-09-15
[Question] How I setup a base path for img or url that use relative path ?


John0King:
I'm new to this project, and I try to render a markdown file through MVC.
I need to config the relative path to a absolute path  such like  `![](image/imageName.jpg)`    => `<img src="/doc/?path=imageName.jpg" />` .
Is there a way to do this ?  Thanks.

xoofx:
Currently, there is no interface to do this automatically. Either you have to process the Markdown AST and fix the links, and then render this AST to HTML. Or make a PR that can introduce this as part of the HTML rendering part, as an option (base url for relative url for example).

markheath:
I've had a crack at doing this in the `HtmlRenderer` class. Should be minimal perf impact if the feature isn't being used


        /// <summary>
        /// Gets a value to use as the base url for all relative links
        /// </summary>
        public Uri BaseUrl { get; set; }

        public HtmlRenderer WriteEscapeUrl(string content)
        {
            if (content == null)
                return this;

            if (BaseUrl != null && !Uri.TryCreate(content, UriKind.Absolute, out Uri _))
            {
                content = new Uri(BaseUrl, content).AbsoluteUri;
            }

            ....


I've made some unit tests too, so if you like this approach I can issue a PR

xoofx:
Yep, looks good for a PR

John0King:
@markheath  this is  great , that works for me !!
but I think maybe it better be a delegate method instead just a property , for example
```c#
public delegate string UrlMarkdownConverter(string rawValue,string extension,bool isRelative)

public UrlMarkdownConverter UrlConverter { get;set; } = (raw,ext,isrelative)=> rawUrl  ; // default value
```

and then I can do this

```C#
reander.UrlConverter = (raw,ext,isRelative)=>
{
    if( isRelative && string.Equals(ext,".md") )
    {
         return raw.Replace(ext,".html"); // better to use Regex class
    }
}
```
to actually generate markdown to HTML files

markheath:
I think if you went that approach it would make more sense to just have a `Func<string,string>` and give complete control to the caller to decide how they want to interpret and transform the URL

John0King:
@markheath   I think  'get extension' and 'the path is relative or not'  may be a very common action, if it not done by `render`,then  it need to be done in your code anyway.  and add those two parameter you still have complete control. 
`Func<string,string>`  is still good for me too .  :P

markheath:
[...]

PR #197, opened 2018-01-17
Convert relative URLs to absolute

markheath:
HTML renderer supports converting relative URLs on links and images to absolute #143

Doesn't touch links inside HTML blocks, just markdown defined blocks

Added a few basic unit tests.


xoofx:
Thanks

PR #199, opened 2018-01-20
Upgrade NUnit3TestAdapter from 3.2 to 3.9 

leotsarev:
to address Resharper test runner problems

The problems with running test in R# is solved by upgrading NUnit3TestAdapter  to latest version.
This is per R# support recommendation and solves issue for me


leotsarev:
For reference https://youtrack.jetbrains.com/issue/RSRP-467835

PR #206, opened 2018-02-06
Added third party extensions to readme.md

macaba:


xoofx:
Thanks, fixed through commit 431fecb

PR #213, opened 2018-02-21
Add alternative `Use` extension method to `MarkdownPipelineBuilder` that receives an object instance

caioproiete:
This PR adds an overload of the `MarkdownPipelineBuilder.Use<T>` method that receives an instance of an `IMarkdownExtension`, for the cases where the extension has a constructor requiring parameters and cannot be instantiated by the existing `Use<T>` method.

Syntax looks like:

```csharp
	.Use(new MyAwesomeExtension("param1", "param2")) // <== this
	.UseAdvancedExtensions()
	.Build();
```

### DIFF
--- a/src/Markdig/MarkdownExtensions.cs
+++ b/src/Markdig/MarkdownExtensions.cs
@@ -52,6 +52,19 @@ namespace Markdig
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


xoofx:
Thanks!

Issue #211, opened 2018-02-20
[Question] `TextWriter.NewLine` is hard-coded to '\n' even when running on Windows


caioproiete:
Hello!

Quick question: Is there a reason to [always uses `\n` for line-breaks instead of `Environment.NewLine`](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Renderers/TextRendererBase.cs#L32), or is just preference?

Markdown files will typically contain fragments of content that - after being converted to HTML - are likely added as content to an HTML template (which itself has line-breaks), which can cause mixed line-endings if the HTML template uses `\r\n`.

Kryptos-FR:
This should probably be an option (with a decent default value, like `Environment.NewLine`) so as to accommodate all scenarios.

caioproiete:
@Kryptos-FR I agree that this should be configurable. I've submitted PR #214 to address that.

I'm curious on why `Environment.NewLine` is not the default, though...

xoofx:
Thanks, PR #214 is merged, will publish a new package soon

PR #214, opened 2018-02-21
Add `ConfigureNewLineExtension` markdown extension

caioproiete:
This PR adds an extension which allows devs to define what string to use for line-endings when writing.

As per my comment in #211, [`Markdig` always uses `\n` for line-breaks](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Renderers/TextRendererBase.cs#L32). This extension exposes an easy way to override the default behavior when defining the pipeline.

Below is an example on how to define that line-endings should use `Environment.NewLine`:

```csharp
	.ConfigureNewLine(Environment.NewLine)
	.UseAdvancedExtensions()
	.Build();
```

---

ps: This PR builds on #213, which should be merged first.

### DIFF
--- a/src/Markdig/MarkdownExtensions.cs
+++ b/src/Markdig/MarkdownExtensions.cs
@@ -29,6 +29,7 @@ using Markdig.Extensions.SmartyPants;
 using Markdig.Extensions.NonAsciiNoEscape;
 using Markdig.Extensions.Tables;
 using Markdig.Extensions.TaskLists;
+using Markdig.Extensions.TextRenderer;
 using Markdig.Extensions.Yaml;
 using Markdig.Parsers;
 using Markdig.Parsers.Inlines;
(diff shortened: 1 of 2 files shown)


src/Markdig/MarkdownExtensions.cs
```diff
@@ -575,5 +589,17 @@ public static MarkdownPipelineBuilder Configure(this MarkdownPipelineBuilder pip
...
+
+        /// <summary>
+        /// Configures the string to be used for line-endings, when writing.
+        /// </summary>
+        /// <param name="pipeline">The pipeline.</param>
+        /// <param name="newLine">The string to be used for line-endings.</param>
+        /// <returns>The modified pipeline</returns>
+        public static MarkdownPipelineBuilder ConfigureNewLine(this MarkdownPipelineBuilder pipeline, string newLine)
+        {
+            pipeline.Use(new ConfigureNewLineExtension(Environment.NewLine));
```
xoofx: This should be `newLine` here instead of `Environment.NewLine`

xoofx:
Thanks! Could you add a few tests?

src/Markdig/MarkdownExtensions.cs
caioproiete: Good catch! At least one of us is awake! :smile:

caioproiete:
>
> Thanks! Could you add a few tests?
>

Tests added via #222

PR #216, opened 2018-03-02
Changed: Configure extensions by reflection

nea:
Hey

This PR is a suggestion for something further in the future.

I changed to the _MarkdownExtensions.Configure()_ method to work with reflection, therefore grabbing all classes implementing the _IMarkdownExtension_ interface and adding them to the pipeline if configured.

I thought of this to be a little more flexible also in the future without having to adapt the switch/case all the time.

But: I know the change is drastic! And it's not 100% backwards compatible to previous _Configure()_ strings as I work with the extension class name which isn't the same as in the switch case for every plugin.

I adapted the tests accordingly and also added a small test for my changes.

I am open for feedback and suggestions as I would be happy to have this system as flexible as possible to use in the future.

Thanks


xoofx:
Thanks for the PR. Indeed, this is a known issue, the reflection approach could look appealing, but there are several issues: 
* not super portable (different scanning behavior across the different .NET framework - some of them giving quite different results, e.g: including user assemblies vs scanning only markdig assembly)
* slow scanning on each markdown pipeline builder ctor
* depends if the assemblies are available - already loaded - at the time the builder is created, so depending on where you call the pipeline builder, you could have different extensions activated in your application, quite unpredictable result
* doesn't allow selective extensions available in case of multiple user custom extensions founded in user assemblies (you get them all, you can't restrict them or select them depending on your requirements)

I would be more in favor of pre-listing the list of registered extensions directly in the pipeline builder, so that you can create your own pipeline builder that adds also a specific set of extensions.

PR #220, opened 2018-03-23
Adding a failing unit test to TestPlainText test class.

hemantkd:
@xoofx Could you confirm if it's a valid test, please?

### DIFF
--- a/src/Markdig.Tests/TestPlainText.cs
+++ b/src/Markdig.Tests/TestPlainText.cs
@@ -13,5 +13,14 @@ namespace Markdig.Tests
             var actual = Markdown.ToPlainText(markdownText);
             Assert.AreEqual(expected, actual);
         }
+
+        [Test]
+        public void TestPlain2()
+        {
+            var markdownText = "*Hello*,\r\n [world](http://example.com)!";
+            var expected = "Hello,\n world!";
+            var actual = Markdown.ToPlainText(markdownText);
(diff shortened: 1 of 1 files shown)


xoofx:
It is failing on appveyor. Don't remember the details of PlainText, but not sure by default it keeps the \n in a paragraph (it requires an option that has not been developed likely)

xoofx:
But not sure to understand, you had this test failing on your machine as well right?

hemantkd:
Yes, it is failing on my machine too.

By writing this test, I wanted to highlight the bug. Once we agree that the expected text in the test is correct then we can discuss about a solution.

When the line slice is empty it is ignored when rendering plain text.

For a List block, a new line is put always before rendering.

I can highlight different scenarios with different tests and work in a TDD fashion but might need some help in understanding the requirements.

hemantkd:
Hi @xoofx,

So, do you think the test is invalid?

xoofx:
No, just closing until there is a PR for the whole problem. Can't merge a test alone that is failing ;)
If you try to fix this, be careful that CommonMark doesn't require newlines and you can break a paragraph with new lines without any problem. That's GitHub that is handling this differently here (and there is an extension in Markdig to do the same). So per se, it is not a bug, but it will require a kind of "extension" system for plaintext.

hemantkd:
Cool, Thanks!

PR #228, opened 2018-05-22
Cross target .NETSTANDARD 2.0

iamcarbon:
Cross target .NETSTANDARD 2.0 to eliminate dependencies on modern runtimes.

### DIFF
--- a/src/Markdig/Markdig.csproj
+++ b/src/Markdig/Markdig.csproj
@@ -51,6 +51,10 @@
     <DefineConstants>$(DefineConstants);NETSTANDARD_11;SUPPORT_UNSAFE</DefineConstants>
   </PropertyGroup>
 
+    <PropertyGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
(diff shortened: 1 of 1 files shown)


src/Markdig/Markdig.csproj
```diff
@@ -7,7 +7,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <VersionPrefix>0.15.0</VersionPrefix>
     <Authors>Alexandre Mutel</Authors>
-    <TargetFrameworks>net35;net40;portable40-net40+sl5+win8+wp8+wpa81;netstandard1.1;uap10.0</TargetFrameworks>
+    <TargetFrameworks>net35;net40;portable40-net40+sl5+win8+wp8+wpa81;netstandard1.1;netstandard2.0;uap10.0</TargetFrameworks>
```
xoofx: A define later in this file should be added at least to add `SUPPORT_UNSAFE`

Could you check also if we can have `SUPPORT_FIXED_STRING` for this target?

iamcarbon: On it.

PR #234, opened 2018-06-28
Fix spelling in package description

ForNeVeR:
Wicktionary [marks](https://en.wiktionary.org/wiki/powerfull) "powerfull" as an "archaic form of powerful", so I suppose it could be fixed?


jwinett:
They say _archaic_, I say _classic_.

xoofx:
Thanks

PR #241, opened 2018-08-29
Expose the number of lines in a document

MihaZupan:
I think this would be a nice feature to have (I know I need it).

I've tried to do it by using a block parser just to count lines, but, while it worked, it seemed overkill considering the value is already known, just hidden.
```csharp
private class LineCounter : BlockParser
{
    public int Count => _count + 1;
    private int _count = -1;
    public override BlockState TryOpen(BlockProcessor processor)
    {
        _count = processor.LineIndex;
        return BlockState.None;
    }
}
```

### DIFF
--- a/src/Markdig/Syntax/MarkdownDocument.cs
+++ b/src/Markdig/Syntax/MarkdownDocument.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Alexandre Mutel. All rights reserved.
+// Copyright (c) Alexandre Mutel. All rights reserved.
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
 namespace Markdig.Syntax
@@ -15,5 +15,10 @@ namespace Markdig.Syntax
         public MarkdownDocument() : base(null)
         {
         }
+
+        /// <summary>
+        /// Gets the number of lines in this <see cref="MarkdownDocument"/>
+        /// </summary>
+        public int LineCount;
     }
 }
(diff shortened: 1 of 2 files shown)


xoofx:
Why not. By curiosity, in which case you are using this?

src/Markdig/Syntax/MarkdownDocument.cs
```diff
@@ -15,5 +15,10 @@ public class MarkdownDocument : ContainerBlock
         public MarkdownDocument() : base(null)
         {
         }
+
+        /// <summary>
+        /// Gets the number of lines in this <see cref="MarkdownDocument"/>
+        /// </summary>
+        public int Lines { get; internal set; }
```
xoofx: Better use `LineCount` instead (Lines can imply that it would return an array). You can also leave it public get set (not internal). It is not something important to protect and doesn't hurt if someone wants to change it (create a MarkdownDocument manually...etc.)

MihaZupan: Commit amended.
As to why, I wrote a tool that checks a bunch of things about markdown 'codebases', and emits warnings. I wrote [my own regex-y parser](https://github.com/MihaZupan/MarkdownValidator/blob/master/src/ReferenceValidator/Parser/MarkdownFile.cs#L88) to detect the structures I needed, but now I decided it would be much easier and much more reliable if I just use a well-established parser.

On another note, I was going to accept a pipeline builder, but now I realise that it's not designed for over multiple files. Should I make users implement some sort of class that returns fresh instances of pipeline/pipeline builder every time? Or is there a cleaner solution I'm missing here?

xoofx:
Thanks!

PR #243, opened 2018-09-11
Fix Link and Footnote reference definitions in syntax tree

MihaZupan:
This does not affect HTML rendering at all, it merely exposes more information to the syntax tree:
- Keep all LinkReferenceDefinitions (doesn't overwrite same-named ones)
- Sets the correct lines and spans for them
- Sets the content and source locations for FootnoteLinkReferenceDefinitions

### DIFF
--- a/src/Markdig.Tests/TestSourcePosition.cs
+++ b/src/Markdig.Tests/TestSourcePosition.cs
@@ -156,11 +156,26 @@ literal      ( 0, 6)  6-7
         }
 
         [Test]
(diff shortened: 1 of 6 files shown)


xoofx:
Thanks for this, looks good, do you think it is complete with enough relevant tests?

MihaZupan:
Yes.
It is further tested in my [(project](https://github.com/MihaZupan/MarkdownValidator), that currently targets the fork instead of NuGet (it relies on these changes).

PR #244, opened 2018-09-17
Expose list item order

MihaZupan:
This makes things like [this](https://github.com/MihaZupan/MarkdownValidator/blob/master/src/MarkdownValidator/Parsing/Parsers/ListBlockParser.cs#L13) possible.


MihaZupan:
Not sure if `Number` would be a more appropriate name than `Order`

xoofx:
> This makes things like this possible.

Thanks. Note that respecting the order is not always an ideal constraint. When you have many items and you don't want to re-order them whenever you insert a value, it is better to use the same number for all the items.

PR #247, opened 2018-09-24
 Expose line start indexes

MihaZupan:
This makes precise source locations much more reliable since it removes the need to ferry around two line and column numbers alongside the span. Sample location algorithm made possible by this:
https://github.com/MihaZupan/MarkdownValidator/blob/master/src/MarkdownValidator/Warnings/WarningLocation.cs#L54-L73
I think it being available only when PreciseSourceLocation is set is fitting.


xoofx:
Thanks! Realized that I didn't push 15.3 while I did the change locally 😅

Issue #252, opened 2018-10-02
Autolink should validate domain name


yufeih:
[GFM Spec](https://github.github.com/gfm/#extended-www-autolink) defines autolink as:

> An extended www autolink will be recognized when the text www. is found **followed by a valid domain**. A valid domain consists of alphanumeric characters, underscores (_), hyphens (-) and periods (.). There must be at least one period, and no underscores may be present in the last two segments of the domain.

> An extended url autolink will be recognised when one of the schemes http://, https://, or ftp://, **followed by a valid domain**, then zero or more non-space non-< characters according to extended autolink path validation

Links like _https://[your-domain]/api_ will not be treated as autolink

https://github.com/dotnet/docfx/issues/3272

PR #253, opened 2018-10-06
Autolink domain name validation

MihaZupan:
#252 

This makes it so AutoLinkParser doesn't match links with domain names, that:
* Have empty segments
* Have less than 2 segments
* Contain invalid characters
* Contain an underscore in the last two segments

This also doesn't match empty emails in mailto: (mailto:@something.com)

@xoofx
I included the Specs.cs file that my VS emits, but it's obviously seriously deformed. Any idea on how to fix it? (I will obviously amend the last commit if I figure that one out)

### DIFF
--- a/src/Markdig.Tests/Specs/AutoLinks.md
+++ b/src/Markdig.Tests/Specs/AutoLinks.md
@@ -1,4 +1,4 @@
-# Extensions
+﻿# Extensions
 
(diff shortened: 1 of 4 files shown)


xoofx:
Thanks! Merged by commit  e390a87 (I tried to rewrote the last Specs commit, without success, but it seems ok, so no worries)

Issue #251, opened 2018-10-01
Empty alt text on reference-style images rendering incorrectly


guypursey:
## Good

This:

`![](image-without-alt-text.jpg)`

Produces this:

```
<p>
  <img src="image-without-alt-text.jpg" alt="">
</p>
```

Which is expected behaviour and good for accessibility.

## Bad

However, if you use a reference style image:

```
![][1]

[1]: image-without-alt-text.jpg
```

You get this:

```
<p>
  <img src="image-without-alt-text.jpg" alt="1">
</p>
```

Which is not good for accessibility -- there's no reason the alt text should be the number `1`.

You can [try this yourself with Markdig here](https://babelmark.github.io/?text=Hello+**Markdig**!%0A%0A!%5B%5D(image-without-alt-text.jpg)%0A%0A!%5B%5D%5B1%5D%0A%0A%5B1%5D%3A+image-without-alt-text.jpg%0A%0A) to verify.

## Summary

I'd expect the reference-style image to behave in the same way as the inline-style image notation which is listed as [example 552 in the CommonMark spec](https://spec.commonmark.org/0.28/#example-552).

But instead it's using the index of the footnote as the alt text.

Expected behaviour renders in most other libraries and crucially [you can see this rendering correctly in CommonMark](https://spec.commonmark.org/dingus/?text=!%5B%5D%5B1%5D%0A%0A%5B1%5D%3A%20(%2Fimage-without-alt-text.jpg)%0A) which Markdig purports to be compliant with.

xoofx:
Fixed by PR #254 by @MihaZupan

guypursey:
Sorry to follow-up on a closed issue, but any idea when the PR above (#254 ) might be included in a release? Thanks.

xoofx:
> Sorry to follow-up on a closed issue, but any idea when the PR above (#254 ) might be included in a release? Thanks.

I don't always push a new release for every PR merged. I wait for a few days/weeks if this is not critical to batch them with other PRs. You can always compile it on your own if you want this immediately.

PR #254, opened 2018-10-07
Empty image alt text for link reference definitions

MihaZupan:
#251 
All tests are passing so I don't think it breaks anything else?


xoofx:
Could you add a test relating to #251? to actually see that this is resolving the original issue (and that we will be able to track a regression there)

xoofx:
Thanks!

Issue #258, opened 2018-10-25
Autolink does not recognize domain names followed by URL fragments query or port


yufeih:
Due to #253, https://github.com#a, https://github.com?a, https://github.com:443 is no longer recognized as autolink. The [spec](https://github.github.com/gfm/#extended-autolink-path-validation) does not say about requiring a `/` after domain name, browsers can also handles these types of links just fine.

cc @MihaZupan

MihaZupan:
I see, I think the only change needed here is to check for these characters along with the '/' check in LinkHelper. I'll do a PR when I get to a computer.

xoofx:
Fixed by PR #260

yufeih:
Thank you guys for the fix, awesome 🎁

PR #260, opened 2018-10-25
Allow #, ? and : instead of only / as the domain separator for AutoLinks

MihaZupan:
Addresses #258 

The issue with the Specs file remains, do you want me to commit it anyway?

### DIFF
--- a/src/Markdig/Helpers/LinkHelper.cs
+++ b/src/Markdig/Helpers/LinkHelper.cs
@@ -690,7 +690,7 @@ namespace Markdig.Helpers
 
                 if (!c.IsAlphaNumeric())
                 {
-                    if (c == '/') // End of domain name
+                    if (c == '/' || c == '?' || c == '#' || c == ':') // End of domain name
                         break;
 
                     if (c == '_')
--- a/src/Markdig.Tests/Specs/AutoLinks.md
+++ b/src/Markdig.Tests/Specs/AutoLinks.md
@@ -1,4 +1,4 @@
-﻿# Extensions
+# Extensions
 
 This section describes the different extensions supported:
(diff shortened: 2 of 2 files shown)


src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -690,7 +690,7 @@ public static bool IsValidDomain(string link, int prefixLength)
 
                 if (!c.IsAlphaNumeric())
                 {
-                    if (c == '/') // End of domain name
+                    if (c == '/' || c == '?' || c == '#' || c == ':') // End of domain name
```
AdmiringWorm: Just an opinion:

this could perhaps also be written as:
```cs
if ("/?#:".Contains(c))
```

This would allow it to be extended to more characters in the future if needed.
Don't know if there is any performance difference though.

MihaZupan: Up to personal taste I guess.
Performance-wise it doesn't really matter in this case, I just try to stick to the same style as the rest of the library.

xoofx: Just made a benchmark comparison for 5 characters in a tests (between contains and ==)

```
   Method |     Mean |     Error |    StdDev |   Median |
--------- |---------:|----------:|----------:|---------:|
 Contains | 641.7 ns | 12.833 ns | 31.959 ns | 629.5 ns |
   Equals | 254.1 ns |  5.030 ns |  8.403 ns | 253.0 ns |
```
And it seems that Equals is still faster. So yeah, let's keep equals in that case

xoofx:
Thanks for this fix!

Issue #261, opened 2018-10-27
It would be nice if autolink could be matched on more leading punctuations


jijiechen:
In current implementation of Markdig as of 0.15.2, links are not automatically created if there are more kinds of leading punctuations than the ones recognized by Markdig at here:

https://github.com/lunet-io/markdig/blob/master/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs#L34

(Only whitespace, `*`, `*`, `_`, `~` `(` are accepted for now).

So if links could be matched and created on more kinds of leading punctuations, that'll be nice. Either list more allowed punctuations or make the `IsValidPreviousCharacter` method an instance method and overridable is helpful. 

I know and fully understand that it's a normal manner to put whitespaces before links in English (because links are just new words in sentences.) But in other languages, like Chinese, it's also normal there are other punctuations before links.

For instance:
```
请访问网址查看更多：https://share.weiyun.com/5D8g7nO
```

Punctuations need to be matched are:
```cs
new []{ '：', '。', '，', '、', '—', '；', '！', '？', '（' };
```

xoofx:
Yes, markdig should follow what [GitHub is actually expecting](https://github.github.com/gfm/#autolinks-extension-) for autolinks.

You can make a PR by changing the test:

```
c.IsWhiteSpaceOrZero() || c == '*' || c == '_' || c == '~' || c == '(';
```

and using a string instead:


```
public string ValidPreciousCharacters {get; set; }

ValidPreciousCharacters = "*_~( \t";

ValidPreciousCharacters.Contains(c)  (check other whitespace characters)
```

PR welcome!

MihaZupan:
Would something like #264 be okay?

(Most of it is just exposing the public field)

jijiechen:
See your just-pushed commit.
I think so, it should match the requirement. Many thanks!

MihaZupan:
Closing as resolved

PR #264, opened 2018-10-29
Make AutoLink ValidPreviousCharacters configurable

MihaZupan:
#261

### DIFF
--- a/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
+++ b/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
@@ -19,7 +19,7 @@ namespace Markdig.Extensions.AutoLinks
         /// <summary>
         /// Initializes a new instance of the <see cref="AutoLinkParser"/> class.
         /// </summary>
-        public AutoLinkParser()
+        public AutoLinkParser(string validPreviousCharacters = DefaultValidPreviousCharacters)
         {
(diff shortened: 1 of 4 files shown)


src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
```diff
@@ -28,20 +28,19 @@ public AutoLinkParser()
...
+        // All such recognized autolinks can only come at the beginning of a line, after whitespace, or any of the delimiting characters *, _, ~, and (.
+        public readonly string ValidPreviousCharacters;
+        public const string DefaultValidPreviousCharacters = "*_~(";
+
         public override bool Match(InlineProcessor processor, ref StringSlice slice)
         {
             // Previous char must be a whitespace or a punctuation
             var previousChar = slice.PeekCharExtra(-1);
-            if (!IsValidPreviousCharacter(previousChar))
+            if (!previousChar.IsWhiteSpaceOrZero() && ValidPreviousCharacters.IndexOf(previousChar) == -1)
```
xoofx: Couldn't we use ValidPreviousCharacters.Contains here instead?

xoofx: Han yes, sorry, actually you are right. IndexOf will be way faster then

xoofx:
Thanks for this!

PR #267, opened 2018-11-09
AutoIdentifierOptions clarification

### LINKED ISSUE #266: Unable to set AutoIdentifierOptions to None / disable
`var pipeline = new MarkdownPipelineBuilder()
            .UseAutoIdentifiers(Markdig.Extensions.AutoIdentifiers.AutoIdentifierOptions.None)
            .Build();`

Does not seem to do anything compared to Default or any of the other options. Maybe something is broken? My use case is that I want to use all of the AdvancedExtensions but disable the auto-identifier option.

redcrayonn:
Clarified 'None' option in the options for the auto-identifiers extension. Fixes #266 .

### DIFF
--- a/src/Markdig/Extensions/AutoIdentifiers/AutoIdentifierOptions.cs
+++ b/src/Markdig/Extensions/AutoIdentifiers/AutoIdentifierOptions.cs
@@ -12,7 +12,7 @@ namespace Markdig.Extensions.AutoIdentifiers
     public enum AutoIdentifierOptions
     {
         /// <summary>
-        /// No options
+        /// No options: does not apply any additional formatting and/or tranformations.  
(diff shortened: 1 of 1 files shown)


src/Markdig/Extensions/AutoIdentifiers/AutoIdentifierOptions.cs
```diff
@@ -12,7 +12,7 @@ namespace Markdig.Extensions.AutoIdentifiers
     public enum AutoIdentifierOptions
     {
         /// <summary>
-        /// No options
+        /// No options: does not apply any additional formatting and/or tranformations.
```
Kryptos-FR: Typo: transformation

PR #280, opened 2019-01-03
Make MediaLinkExtension XML well-formed

FranklinWhale:
Change `<video controls>` to `<video controls="">` and `<iframe allowfullscreen>` to `<iframe allowfullscreen="">`

### DIFF
--- a/src/Markdig.Tests/Specs/MediaSpecs.md
+++ b/src/Markdig.Tests/Specs/MediaSpecs.md
@@ -17,9 +17,9 @@ Allows to embed audio/video links to popular website:
 
 ![Video5](https://ok.ru/video/26870090463)
 .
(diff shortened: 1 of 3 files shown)


xoofx:
Hm, allowfullscreen should be well formed for HTML5 at least. Do you expect to target something else?

FranklinWhale:
It is valid HTML 5, but it is not valid XHTML 5 because of the omission of the attribute value assignment part, making the syntax not XML well-formed. Adding back the value assignment makes it valid for both HTML and XHTML.

https://www.w3.org/TR/html/infrastructure.html#boolean-attribute

FranklinWhale:
Also, if the content is well formed XML, it can processed easily using the LINQ to XML API.

FranklinWhale:
Closing this in favour of #281

PR #281, opened 2019-01-03
Render XML well-formed boolean attribute

FranklinWhale:
Although HTML boolean attribute can have its value assignment omitted, such omission is not supported in XHTML because the syntax is not XML well-formed. Being well-formed XML allows the content to be processed by XML API such as LINQ to XML. This PR changes the attribute rendering method so that the value assignment of a boolean attribute is always specified (https://www.w3.org/TR/html/infrastructure.html#boolean-attribute).

This PR, if approved, replaces #280.


src/Markdig/Renderers/HtmlRenderer.cs
```diff
@@ -314,12 +314,9 @@ public HtmlRenderer WriteAttributes(HtmlAttributes attributes, Func<string, stri
                 foreach (var property in attributes.Properties)
                 {
                     Write(" ").Write(property.Key);
-                    if (property.Value != null)
```
xoofx: Maybe I would like to make this whole thing a bit more configurable instead of taking the least common denominator.

Could you add an enum HtmlMode enum with HTML5, XHTML5 maybe (or whatever values would make more sense)? and if/else according to it where it makes sense?

FranklinWhale: If such option exists, should it be a static option applying all instances of `HtmlRenderer`?

That allows an application using the library to have XML well-formed attribute across all extensions.

Personally I think the option is not necessary because the revised output remains valid HTML.

xoofx: no static, instance variable (configurable via an extension)

FranklinWhale: Actually I have been using the `Markdown.ToHtml` method to convert Markdown to HTML, which has been producing well-formed XML, until `UseMediaLinks()` is added to the pipeline.

Originally I thought it is an issue related to the Media Link Extension. However, on second thought, I think it is the behaviour of the renderer that should be changed and thus I made this PR. Since the HTML Specification defines a way to specify Boolean attribute that is both valid HTML and XML, it doesn't seem necessary to support the omitted form.

The Specification specifically stated that

> A boolean attribute without a value assigned to it (e.g. checked) is implicitly equivalent to one that has the empty string assigned to it (i.e. checked="").

So, the empty string assignment is added when the attribute value is null.

FranklinWhale: I looked for all current use of `AddPropertyIfNotExist` and `AddProperty` and found that only the Media Link Extension is passing `null` to attribute value. However, if the Generic Attributes Extension is enabled, Boolean attribute may be passed into the attribute collection. The attribute rendering method appears the best place to resolve the well-formness issue.

FranklinWhale: @xoofx Are you expecting me to do the following instead of `Markdown.ToHtml` after the enum option is added?

```cs
private static string ConvertMarkdownToHtml(string markdown, MarkdownPipeline pipeline = null)
{
	if (markdown == null) throw new ArgumentNullException(nameof(markdown));
	var writer = new StringWriter();
	Markdown.Convert(markdown, new HtmlRenderer(writer) { BooleanAttributeMode = BooleanAttributeMode.EmptyString }, pipeline);
	return writer.ToString();
}
```

xoofx: no that's fine, don't add an attribute, this is not critical for the lib and if the output is compatible with both HTML5 and XHTML5, then ok.

Are you done with the commits for this PR?

FranklinWhale: Yes. No further change is expected. Thanks.

xoofx:
Thanks!

Issue #282, opened 2019-01-07
Allow more leading # in ATX heading as a customization option


yufeih:
We are using `TOC.md` to represent table of contents, It is a restricted markdown format that only allows several known markdown syntax, like links and ATX headers.

Sometimes these files can be very big and exceed [maximum leading count](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Parsers/HeadingBlockParser.cs#L54) as defined by common mark spec, like [this one](https://raw.githubusercontent.com/MicrosoftDocs/VBA-Docs/master/api/TOC.md),

The ask is if it is possible to have an option in `HeadingBlockParser` that can override the default max leading count (6), so we can reuse `HeadingBlockParser` to parse our special TOC markdown.

Related https://github.com/dotnet/docfx/issues/3874

xoofx:
Sure, via an extension and a slight change on `HeadingBlockParser`, PR welcome

OsmondJiang:
Thanks a lot for quickly merging the PR 👍, it would be great if we can know when we can get a new nuget version?

xoofx:
It should be come with upcoming `0.15.7` in a few minutes

PR #283, opened 2019-01-07
make the max leading # count in heading block parser configurable

OsmondJiang:
Will implement the extension part in docfx side
#282


xoofx:
Thanks!

PR #285, opened 2019-01-12
Adds markdown parser context

### LINKED ISSUE #284: Any way of accessing (external) contextual data in a parser?
I've written a MarkDig parser and in the parser I would want to get access to some contextual information about the document being parsed (that's not available in the document itself).

Is there any way of passing contextual data to a parser on a parse basis (just like I can in Scriban when rendering)?

```csharp
// Something like this?
var html = Markdown.ToHtml(markdown, _pipeline, context);
```

If there is no work around for my problem, I would be happy to provide a pull request for this functionality if you think it would be a good fit for MarkDig.

patriksvensson:
Closes #284

### DIFF
--- a/src/Markdig/MarkdownParserContext.cs
+++ b/src/Markdig/MarkdownParserContext.cs
@@ -10,7 +10,7 @@ namespace Markdig
         /// <summary>
         /// Gets or sets the context property collection.
         /// </summary>
-        public IDictionary<object, object> Properties { get; set; }
+        public Dictionary<object, object> Properties { get; }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="MarkdownParserContext" /> class.
--- a/src/Markdig/MarkdownParserContext.cs
+++ b/src/Markdig/MarkdownParserContext.cs
@@ -1,0 +1,23 @@
+using System.Collections.Generic;
+
+namespace Markdig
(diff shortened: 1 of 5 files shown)


patriksvensson:
I've added the new context parameter to the methods I thought made sense in `Markdown.cs`, but would want some feedback on whether or not this is correct or not.

src/Markdig/MarkdownParserContext.cs
```diff
@@ -0,0 +1,23 @@
...
+{
+    /// <summary>
+    /// Provides a context that can be used as part of parsing Markdown documents.
+    /// </summary>
+    public sealed class MarkdownParserContext
+    {
+        /// <summary>
+        /// Gets or sets the context property collection.
+        /// </summary>
+        public IDictionary<object, object> Properties { get; set; }
```
xoofx: You can expose only a `Dictionary<object, object>`, not a `IDictionary` and remove also the `setter`.

xoofx: Nevermind, forgot that I can do that right away from github, so simple 😅

xoofx:
Thanks!

sveins12:
How can you access the Properties dictionary from within a markdown extension?

PR #288, opened 2019-01-18
Rewrite Descendants iteratively

MihaZupan:
I rewrote the `Descendants` extension methods to eliminate recursion. **The order of returned elements is kept the same.**

This comes with a significant memory allocation reduction when using `Descendants` over the entire AST like I do in my project.

Benchmarking between the recursive and iterative approaches I get this:

Iterative:

|              Method |     Mean | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|-------------------- |---------:|------------:|------------:|------------:|--------------------:|
|             Markdig | 212.9 ms |  99333.3333 |           - |           - |            74.89 MB |
| Markdig_Descendants | 228.8 ms |  98000.0000 |   2000.0000 |           - |            78.22 MB |

Recursive:

|              Method |     Mean | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|-------------------- |---------:|------------:|------------:|------------:|--------------------:|
|             Markdig | 215.8 ms |  99333.3333 |           - |           - |            74.89 MB |
| Markdig_Descendants | 242.9 ms |  98000.0000 | 119000.0000 |   6000.0000 |            104.3 MB |

Where `Markdig` is only parsing to the AST while `Markdig_Descendants` parses to AST and loops over descendants once.

I have tested that the order is not modified when calling descending anywhere in the test suite.

I don't have tests for `Descendants<T>` of `ContainerInline `and `ContainerBlock `as they appear to not be used anywhere in Markdig itself, but I have included iterative implementations for those as well.

In the future, a few bytes could be shaved off by using ValueTuple instead of two stacks.

### DIFF
--- a/src/Markdig/Syntax/MarkdownObjectExtensions.cs
+++ b/src/Markdig/Syntax/MarkdownObjectExtensions.cs
@@ -89,8 +89,20 @@ namespace Markdig.Syntax
         /// </returns>
         public static IEnumerable<T> Descendants<T>(this ContainerBlock block) where T : Block
         {
-            foreach (var subBlock in block)
+            // Fast-path an empty container to avoid allocating a Stack
+            if (block.Count == 0) yield break;
+
+            Stack<Block> stack = new Stack<Block>();
+
+            int childrenCount = block.Count;
+            while (childrenCount-- > 0)
             {
+                stack.Push(block[childrenCount]);
+            }
+
+            while (stack.Count > 0)
+            {
+                var subBlock = stack.Pop();
                 if (subBlock is T subBlockT)
                 {
                     yield return subBlockT;
@@ -98,9 +110,10 @@ namespace Markdig.Syntax
 
                 if (subBlock is ContainerBlock subBlockContainer)
                 {
-                    foreach (var sub in subBlockContainer.Descendants<T>())
+                    childrenCount = subBlockContainer.Count;
+                    while (childrenCount-- > 0)
                     {
-                        yield return sub;
+                        stack.Push(subBlockContainer[childrenCount]);
                     }
                 }
             }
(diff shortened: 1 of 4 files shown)


src/Markdig/Syntax/MarkdownObjectExtensions.cs
```diff
@@ -1,131 +1,119 @@
...
+        /// </summary>
+        /// <param name="markdownObject">The markdown object.</param>
+        /// <returns>An iteration over the descendant elements</returns>
+        public static IEnumerable<MarkdownObject> Descendants(this MarkdownObject markdownObject)
+        {
+            // Fast-path an object with no children to avoid allocating Stack objects
+            if (!(markdownObject is ContainerBlock) && !(markdownObject is ContainerInline)) yield break;
+
+            // TODO: A single Stack<(MarkdownObject block, bool push)> when ValueTuples are available
+            Stack<MarkdownObject> stack = new Stack<MarkdownObject>();
```
xoofx: You can use instead a small internal struct for that, that's fine

xoofx:
Yeah, I remember that I was very lazy for this one, so I'm glad that you are optimizing this. Indeed, that's much better 👍

src/Markdig/Syntax/MarkdownObjectExtensions.cs
MihaZupan: Testing with a struct in a single stack I go from a 3.3 to a 4.5 MB increase in allocations. Want me to commit it anyway?

MihaZupan:
I've tried to rewrite it before for fun, but failed at keeping the order (didn't think to keep the extra bool).

I was just surprised to see this method show up when doing some profiling yesterday

PR #291, opened 2019-01-19
Allow single character abbreviations

MihaZupan:


src/Markdig.Tests/Specs/AbbreviationSpecs.md
```diff
@@ -97,4 +97,13 @@ Abbreviations should only match when surrounded by whitespace:
 PRAA
 .
 <p>PRAA</p>
+````````````````````````````````
+
+One-character abbreviations should be allowed
```
xoofx: You need to save the `Specs.tt` file from VS in order to regenerate the tests (same for #292)

MihaZupan: I know, but didn't want to commit 20k lines of change again. A lot of new lines appear whenever I regenerate the spec.

xoofx: It is fine, otherwise the tests are not running. I don't check the changes in this file

MihaZupan:
Closing as related to #294

PR #294, opened 2019-01-21
Alternative specs generation

MihaZupan:
This generates test csharp similar to Specs.tt with a few differences:
- Tests from different specs are separated into different files (see https://github.com/lunet-io/markdig/commit/ffdfd427ced1c94b40b630bfac4b6bed6a460ee9) This makes it easier to see what changed in diffs
- They are also in separate namespaces (makes navigating tests from TestViewer easier)
- Removed partial class for TestFixtures (now tests are grouped into a single class - style change)

[Parsing C# code](https://github.com/lunet-io/markdig/blob/7887cf7348014b38b3f9c19b8ed38bcde9440bd5/src/SpecFileGen/Program.cs)

Comments and tests are kept the same.
Since this removes template generation, it means generation can be invoked during CI and the build failing if specs files are modified but not regenerated:
https://github.com/lunet-io/markdig/blob/dd683512357006f43ab193dfc49166fcd53c7467/appveyor.yml#L31-L32
https://github.com/lunet-io/markdig/blob/dd683512357006f43ab193dfc49166fcd53c7467/src/SpecFileGen/Program.cs#L109-L113

When running the SpecGen, files will only be modified if there are any changes (so as not to bump Generated timestamps for all files).

Addressing #192, including CommonMark.md as one of the spec files seems appropriate since the spec doesn't change versions too often.


xoofx:
LGTM, btw as part of this, could you add a test that is checking that none of the md files is more recent than the currently `Markdig.Tests.` running assembly? (and log an error specifying that SpecFileGen should be run)

MihaZupan:
Currently it will generate all spec files, compare them with the version on disk, and replace them if changed.
If any spec is changed and the `CI` environment variable is set it will exit with code 1 and emit an error message.

xoofx:
> Currently it will generate all spec files, compare them with the version on disk, and replace them if changed.
If any spec is changed and the CI environment variable is set it will exit with code 1 and emit an error message.

But only when running the tests in the CI, otherwise a user making a local change and running the tests will not see that he forgot to run SpecFileGen, right?. 

So by just adding a test that checks that md files are not more recent, you get at least a local error (and I would prefer to avoid the SpecFileGen to run on every tests locally)

MihaZupan:
Oh I see, yes that'd be better.
I'll add the test

MihaZupan:
One downside to this is if a user modifies a specification file but reverts the changes, the test will still fail.

xoofx:
> One downside to this is if a user modifies a specification file but reverts the changes, the test will still fail.
Other than checking if SpecFileGen emits a different spec in such scenario, I can't think of a workaround.

Yeah, but with a proper source control revert, the timestamp should be preserved at least

> One other way would be to not have SpecFileName.cs files at all, but have one of the tests parse the Specification files and run tests.

I didn't want to do that back in the days where ReSharper/NUnit was a bit clunky working with .NET Core projects... you can try, it would simplify the overall setup (no cs files anymore)

MihaZupan:
I could do it, but I think that the drawbacks of not having a test explorer overview and simple per-test debugging is not worth it.

xoofx:
> I could do it, but I think that the drawbacks of not having a test explorer overview and simple per-test debugging is not worth it.

Ok, so let's stick with the check with md timestamp against assembly, this should cover the most common case.

MihaZupan:
In that case I believe the PR's ready

xoofx:
Thanks for this!

PR #295, opened 2019-01-22
Add the license expression to the `csproj`

TravisEz13:
Chose the license expression based on what GitHub Detects.

A project I work on generates our Third Party notices files using ClearlyDefined.io and they are having trouble finding you license.  They said this would fix that.  This assumes you package based on the csproj not a nuspec that I couldn't find.


xoofx:
Thanks!

PR #301, opened 2019-01-29
Improved emphasis parser

MihaZupan:
This allows for emphasis parsing with min/max above 2 characters.

I marked `IsDouble` on `EmphasisInline` as Obsolete as described below. This should not be a breaking change for existing parsers/renderers.
https://github.com/lunet-io/markdig/blob/d232e847dd1a6822d72cb09740511ee84729100a/src/Markdig/Syntax/Inlines/EmphasisInline.cs#L24-L36

I also [changed](https://github.com/lunet-io/markdig/pull/301/files#diff-b5d0fcbc89fb349b18e37bbef25bdeb4L29) the way adding a `CreateEmphasisInline` works (since there can no longer be a `bool isStrong` identifier), so now there's no need to save the previous one, attempt the new one and revert to the old one on callsites, instead adding the delegate to the list, returning null if not providing an emphasis. `CreateEmphasisInlineDelegate` is marked as obsolete in favour of `TryCreateEmphasisInlineDelegate`. Note that this **shouldn't** be a breaking change, since for now the parser will attempt to use the old delegate first.
https://github.com/lunet-io/markdig/blob/42b201d8c2f599b9f3a2f4ba6f89a412be45d745/src/Markdig/Extensions/CustomContainers/CustomContainerExtension.cs#L29-L36


src/Markdig/Helpers/LegacyHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+using System;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace Markdig.Helpers
+{
+    internal static class LegacyHelper
+    {
+        [MethodImpl(MethodImplOptionPortable.AggressiveInlining)]
+        public static bool HasFlag(this Enum enumeration, Enum flags)
```
xoofx: Couldn't we use instead the new generic constraint for that? (`where TEnum : System.Enum`, likely requiring the project to use C#7.3)

xoofx:
LGTM

src/Markdig/Helpers/LegacyHelper.cs
MihaZupan: I did not know about this addition.
Added the generic constraint now

MihaZupan: I just added the helper method for net35 since I find the old way of checking for flags unintuitive.

src/Markdig/Helpers/LegacyHelper.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace Markdig.Helpers
+{
+    internal static class LegacyHelper
+    {
+        [MethodImpl(MethodImplOptionPortable.AggressiveInlining)]
+        public static bool HasFlag<TEnum>(this TEnum enumeration, TEnum flags) where TEnum : Enum
+        {
+            Debug.Assert(enumeration.GetType() == flags.GetType());
+            ulong enumVal = Convert.ToUInt64(enumeration);
```
xoofx: actually, looking at the code behind, Convert.ToUInt64 typically incurs an allocation (cast IConvertible) which is just super bad. I would prefer that we remove this method entirely. It's not that annoying in the codebase.

MihaZupan: Afaik there's no way to see how many people are using the NuGet by framework version right?

xoofx: no unfortunately

PR #304, opened 2019-02-05
typo: dib -> div

nschonni:


nschonni:
Closing and reopening to see if CI goes green

MihaZupan:
You modified the specification file, the CI is complaining that the generated tests don't match. Running SpecFileGen should fix it.
(tho the output file names changed in a recent commit so it would have to be done against master)

nschonni:
@MihaZupan thanks for the pointer! Got it green again 😄

PR #309, opened 2019-02-20
remove lazy initialize

yishengjin1413:
Fix #308


yishengjin1413:
@xoofx would you help to release a new version for us to reference if the change is ok for you?

PR #310, opened 2019-02-21
Improve CompactPrefixTree's memory footprint

MihaZupan:
Drops CompactPrefixTree's allocations to about half. As a result build times are also lower by about 50%.


xoofx:
Thanks!

PR #314, opened 2019-02-28
Inline code block fix for script ref

mbhattacharya-rythmos:
Inline code block fix for script ref inside table


MihaZupan:
Wtf is this

xoofx:
@mbhattacharya-rythmos I guess it is not meant to this repository but likely to a Unity one right? 😅

PR #315, opened 2019-03-07
Fix ToPlainText with htmlentity

OpportunityLiu:
The `ToPlainText` should not resulted in html entities.

Current version:
```cs
> Markdig.Markdown.ToPlainText("foo<bar")
"foo&lt;bar\n"
> Markdig.Markdown.ToPlainText("foo&lt;bar")
"foo&lt;bar\n"
```

Expacted:
```cs
> Markdig.Markdown.ToPlainText("foo<bar")
"foo<bar\n"
> Markdig.Markdown.ToPlainText("foo&lt;bar")
"foo<bar\n"
```


xoofx:
Thanks!

PR #318, opened 2019-03-08
fix #242: HtmlCustomContainerRenderer does not honour EnableHtmlForBlock == false

### LINKED ISSUE #242: HtmlCustomContainerRenderer does not honour EnableHtmlForBlock == false
We're building our pipeline with the `UseAdvancedExtensions` method as part of it and using this pipeline for rendering the same piece of text as both HTML and plaintext. For the plaintext we are setting up a new HtmlRenderer as:

```
 var renderer = new HtmlRenderer(writer)
                {
                    EnableHtmlForBlock = false,
                    EnableHtmlForInline = false
                };
```

However any content that 'HtmlCustomContainerRenderer' would pick up (i.e. between nested sets of 3 or more colons) is still rendered out as HTML. Having dug into the source I can see that 'HtmlCustomContainerRenderer' does not honour the 'EnableHtmlForBlock' bool that is set on the renderer, which every other extension we use does.

Any chance of this being honoured in a future release?

OpportunityLiu:
See #242

### DIFF
--- a/src/Markdig.Tests/TestParser.cs
+++ b/src/Markdig.Tests/TestParser.cs
@@ -143,7 +143,7 @@ $\frac{n!}{k!(n-k)!} = \binom{n}{k}$
             var html = Markdown.ToHtml(math, pl);
             Console.WriteLine(html);
 
-            Assert.IsTrue(html.Contains("<p><span class=\"math\">\\("),"Leading bracket missing");
+            Assert.IsTrue(html.Contains("<p><span class=\"math\">\\("), "Leading bracket missing");
             Assert.IsTrue(html.Contains("\\)</span></p>"), "Trailing bracket missing");
         }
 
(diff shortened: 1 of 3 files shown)


OpportunityLiu:
Should methods like `WriteOpeningTag(string tag, HtmlAttributes attributes, Func<string, string> classFilter = null)`, `WriteClosingTag(string tag)` and `WriteSelfClosingTag(string tag, HtmlAttributes attributes, Func<string, string> classFilter = null)` be added to `HtmlRender` and avoid write html tags directly by `HtmlObjectRenderer`s? Many extensions have same problem.

xoofx:
Maybe, but I'm worried that there are cases in the code base where this translation would not be that easy

OpportunityLiu:
Well, I'll have a try.

PR #319, opened 2019-03-11
Allow unicode domain names

### LINKED ISSUE #316: No-ascii characters should be supported in AutoLinkInline
https://babelmark.github.io/?text=%3Chttp%3A%2F%2F%E2%98%83.net%3F%E2%98%83%3E

Input:
```md
<http://☃.net?☃>
```

Expacted:
```html
<p>
  <a href="http://xn--n3h.net?%E2%98%83">http://☃.net?☃</a>
</p>
```

MihaZupan:
Fixes #316 

[`IDNA encoding`](https://en.wikipedia.org/wiki/Internationalized_domain_name#Example_of_IDNA_encoding) is not used on `standard1.1` and `portable` as [explained here](https://github.com/lunet-io/markdig/issues/316#issuecomment-471211325).


xoofx:
Thanks! I think we should drop support for netstandard1.1 then and switch to the minimum netstandard version supporting this.

MihaZupan:
Portable as well or only std1.1?

xoofx:
yeah, we can remove portable as well.

xoofx:
btw, I wanted to add you as a collaborator of markdig?  (instead of a contributor as it is today) you will have a bit more access to the project (moderate issues...etc.), as you have been quite involved, I was wondering if you would be interested?

MihaZupan:
Of course, thank you :)

xoofx:
This line can be likely removed:

https://github.com/lunet-io/markdig/blob/f8df3c80e16e98ffe0a0ce8a9a5b7b0f4c345b2f/src/Markdig/Markdown.cs#L20

Haven't checked further, but you can double check (I just did a search on GitHub search, might not be complete!)

MihaZupan:
I tried, looks like uap10 falls under netstandard11 as far as that check is concerned and it doesn't support the alternative

xoofx:
> I tried, looks like uap10 falls under netstandard11 as far as that check is concerned and it doesn't support the alternative

Han ok

xoofx:
Oh, also, as I forgot whenever I have to compile the `changelog.md`, can you add a new line for the next version about this fix in `changelog.md` (with a WIP on the version so that we know what to integrate)

PR #327, opened 2019-04-05
CommonMark 0.29, DisableHeadings extension, AutoLinkOptions

### LINKED ISSUE #83: Removing of HeadingBlockParser from pipeline.BlockParsers is not affect setext headings
I'm trying to make stop markdig from parsing headers (h1-h6).

```
        public static MarkdownPipelineBuilder DisableHeadings(this MarkdownPipelineBuilder pipeline)
        {
            var headingBlockParser = pipeline.BlockParsers.Find<HeadingBlockParser>();
            if (headingBlockParser != null)
            {
                pipeline.BlockParsers.Remove(headingBlockParser);
            }
            return pipeline;
        }
```
But this code doesn't affect [setext headings](https://github.com/lunet-io/markdig/blob/4f52e893eef8401ad7e50f7442c3dc01996f3e91/src/Markdig.Benchmarks/spec.md#setext-headings).
Is it normal behavior? How can I solve this problem?
Thanks.

### LINKED ISSUE #145: AutoLinks option to open in new window
Hi,

I noticed that JiraLinks has this feature, but not AutoLinks. Would you consider a PR for this?

Thanks.

MihaZupan:
Closes #83 
Closes #145 

Adds Spec file system for Normalize and PlainText renderers
Adds `DisableHeadings()` extension method on the pipeline builder to disable ATX and Setex headings.
Adds `AutoLinkOptions` with `OpenInNewWindow` and `UseHttpsForWWWLinks`.


src/Markdig.Tests/PlainTextSpecs/SamplePlainText.cs
```diff
@@ -0,0 +1,35 @@
+// Generated: 2019-03-17 11:31:23
```
xoofx: I think that we should start to use the `.generated.cs` postfix for these files, as it is becoming confusing to have multiple of them around and not being sure if they are generated or not.

src/Markdig/Extensions/AutoLinks/AutoLinkOptions.cs
```diff
@@ -0,0 +1,22 @@
+namespace Markdig.Extensions.AutoLinks
+{
+    public class AutoLinkOptions
+    {
+        public AutoLinkOptions()
+        {
+            ValidPreviousCharacters = "*_~(";
+        }
+
+        public string ValidPreviousCharacters { get; set; }
```
xoofx: I'm a bit worried to introduce a mutable option class here, as a pipeline is supposed to be immutable after a build...
I don't recall if we have many cases already like this in the code base, but we should ideally strive for making these classes immutable

MihaZupan: Looking at other extensions, `PipeTable`, `Media` and `JiraLink` have mutable option classes while `SmartyPantOptions` is unmodifiable.

Would you rather these all have readonly fields and extra constructor params?

I can't really think of a clean way to enforce immutability otherwise (unless extensions made an internal copy and changes wouldn't propagate after the initial setup)

xoofx: yeah, C# is really poorly helping us here unfortunately... ok let it as it is for this one

We will have to later decide what to do with all these mutable classes and if we want to break them

MihaZupan: A property like https://github.com/lunet-io/markdig/pull/327/commits/befd1ca8464a7cc4ac5d0957f47ee726c8286c2d#diff-b59084d198ccb5dc29ab3f390ae3fcc7R15 falls under the same category them

xoofx: yes, as stated in the original issue #5 about immutable, these are open problems...

src/Markdig/Parsers/FencedBlockParserBase.cs
```diff
@@ -1,231 +1,241 @@
...
-                return BlockState.BreakDiscard;
-            }
-
-            // Reset the indentation to the column before the indent
-            processor.GoToColumn(processor.ColumnBeforeIndent);
-
-            return BlockState.Continue;
-        }
-    }
+// Copyright (c) Alexandre Mutel. All rights reserved.
```
xoofx: Weird, why this file is being marked as completely modified?

xoofx:
LGTM, thanks @MihaZupan for working so quickly on the 0.29 specs changes!

src/Markdig/Parsers/FencedBlockParserBase.cs
MihaZupan: MB, apparently I let VS set \r\n line endings for this file.
Actual changes: [gist diff](https://gist.github.com/MihaZupan/2457ea2497f4e58fdd5616a8a75900bf/revisions#diff-44589aff4202ecb8017748d2d2e7a181)

src/Markdig/Parsers/FencedBlockParserBase.cs
```diff
@@ -1,231 +1,241 @@
...
+            string argString = null;
+
+            // An info string cannot contain any backsticks (unless it is a tilde block)
+            int firstSpace = -1;
+            if (openingCharacter == '`')
+            {
+                for (int i = line.Start; i <= line.End; i++)
+                {
+                    char c = line.Text[i];
+                    if (c == '`')
```
xoofx: Shouldn't this using `openingCharacter ` instead here?

MihaZupan: This checks that there are no backtick characters in the info string if the opening character is also a backtick, as from 0.29 backticks in info strings are allowed for tilde fenced blocks.

xoofx: oops sorry, yes, let me check again the file (messy diff in github)

Issue #306, opened 2019-02-09
Markdown.ToHtml takes more than two minutes to complete when processing the 32K file


Metalnem:
**Markdown.ToHtml** takes more than two minutes to complete when processing the 32K file from the [attached archive](https://github.com/lunet-io/markdig/files/2847748/Test.zip). You can reproduce this by running the following program and passing it the path to the extracted file as a command line argument:

```csharp
using System.IO;

namespace Markdig.Run
{
  public class Program
  {
    public static void Main(string[] args)
    {
      var text = File.ReadAllText(args[0]);
      var pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().Build();
      Markdown.ToHtml(text, pipeline);
    }
  }
}
```

I'm using Markdig 0.15.7 and .NET Core 2.2.103.

Found via [SharpFuzz](https://github.com/Metalnem/sharpfuzz).

MihaZupan:
This does not appear to be an issue on the master branch. There is a slight difference in output
https://gist.github.com/MihaZupan/a5ce072fffec28315cbdcfd779b203e9/revisions#diff-20e4e18dfae3d1b38df3bce24863b74dR3
See the end of the huge block

MihaZupan:
I believe this was addressed by the changes to the Abbreviation parser.

On 15.7, [this loop](https://github.com/lunet-io/markdig/blob/master/src/Markdig/Parsers/InlineProcessor.cs#L199-L281) seems to iterate over every character

Metalnem:
You are right, the master works fine. Do you want me to close this issue, or is there something that can be done about the O(n^2) complexity?

PR #359, opened 2019-07-28
[WIP] Optimize abbreviation parser re-entry

MihaZupan:
This improves performance in cases like #306 where the AbbreviationParser's PostMatch was running over the entire LiteralInline, every time a new character was appended to it. Now it only checks the ending.


xoofx:
@MihaZupan it is marked as WIP but is it actually fine to merge it or?

MihaZupan:
It currently only addresses a special edge case such as the input from #306. I will have to look into it a bit more for the general case, no need to merge it right now.

MihaZupan:
Closing until I get back to this

PR #377, opened 2019-10-14
10% time and 50% memory improvement

MihaZupan:
A bunch of random optimizations.

Most interesting (and responsible for perf) ones to look at are:
253be5c afe4308 aefad21

I was mainly playing whack-a-mole with memory allocations, time improvement comes mainly from 253be5c.

### DIFF
--- a/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
+++ b/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
@@ -254,7 +254,7 @@ namespace Markdig.Extensions.AutoLinks
             return countBrackets <= 0;
         }
 
-        private class ListOfCharCache : DefaultObjectCache<List<char>>
+        private sealed class ListOfCharCache : DefaultObjectCache<List<char>>
         {
             protected override void Reset(List<char> instance)
             {
--- a/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
+++ b/src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
@@ -49,150 +49,152 @@ namespace Markdig.Extensions.AutoLinks
             }
 
             List<char> pendingEmphasis = _listOfCharCache.Get();
-            // Check that an autolink is possible in the current context
-            if (!IsAutoLinkValidInCurrentContext(processor, pendingEmphasis))
+            try
             {
-                goto ReturnFalse;
-            }
-
-            var startPosition = slice.Start;
-            int domainOffset = 0;
-
-            var c = slice.CurrentChar;
-            // Precheck URL
-            switch (c)
-            {
-                case 'h':
-                    if (slice.MatchLowercase("ttp://", 1))
-                    {
-                        domainOffset = 7; // http://
-                    }
-                    else if (slice.MatchLowercase("ttps://", 1))
-                    {
-                        domainOffset = 8; // https://
-                    }
-                    else goto ReturnFalse;
-                    break;
-                case 'f':
-                    if (!slice.MatchLowercase("tp://", 1))
-                    {
-                        goto ReturnFalse;
-                    }
-                    domainOffset = 6; // ftp://
-                    break;
-                case 'm':
-                    if (!slice.MatchLowercase("ailto:", 1))
-                    {
-                        goto ReturnFalse;
-                    }
-                    break;
-
-                case 'w':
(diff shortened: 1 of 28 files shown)


src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
```diff
@@ -182,21 +185,23 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             if (Options.OpenInNewWindow)
             {
                 inline.GetAttributes().AddPropertyIfNotExist("target", "blank");
-            }
-
-            return true;
+            }
+
+            _listOfCharCache.Release(pendingEmphasis);
```
xoofx: Wondering if a plain try/finally wouldn't be easier in that case and avoid the goto?

src/Markdig/Helpers/CustomArrayPool.cs
```diff
@@ -0,0 +1,86 @@
...
+            public Bucket(int numberOfBuffers)
+            {
+                _buffers = new T[numberOfBuffers][];
+            }
+
+            public T[] Rent()
+            {
+                T[][] buffers = _buffers;
+                T[] buffer = null;
+                if (Interlocked.CompareExchange(ref _lock, 1, 0) == 0)
```
xoofx: Are you sure that this code is thread safe?
If we have 2 threads entering this method at the exact same time, thread-1 enter this line (so _lock is then set to 1), so thread-2 can't enter (_lock is 1), so it goes directly to `return buffer` which will be null, no?

xoofx: (this case is probably a better fit for a slim mutex)

src/Markdig/Extensions/AutoLinks/AutoLinkParser.cs
MihaZupan: Haven't thought of that one. I suppose it would be cleaner since there are many exit points. I'll make the change.

src/Markdig/Helpers/CustomArrayPool.cs
MihaZupan: This is heavily adapted from the [CoreFx's ConfigurableArrayPool](https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/Buffers/ConfigurableArrayPool.cs#L163). A SpinLock is used there instead.

I expect very little contention on this lock. Parsing is mostly done by one thread and even then dealing with the StringLine buffers all that common.
As you said, in this implementation if there is contention, the second thread will return null. That will fallback to a regular `new T[length]` allocation. Such a buffer can still be returned to the pool later on.

The reason I opted for that is because not all buffers will get released. In the case of StringLineGroup, only objects where inlines are processed will release the buffer in the end. In other cases the Lines are still available on the AST and adding a finalizer to reclaim them is way to expensive.

As such only buffers from objects that have their ProcessInlines set will get reclaimed. For other objects, buffers will only be returned when resizing, the final buffer on them will be reclaimed by the GC eventually and not returned to the pool.

xoofx: oh, ok, sorry, didn't look further with the `return SelectBucket(length)?.Rent() ?? new T[length];`
yeah that's ok then

MihaZupan: In other words, buffers aren't guaranteed to be returned from the actual bucket, they can also be allocated on the spot if the bucket is contended/at capacity/of incorrect size.
Over-allocating here is fine since some buffers will never get returned anyway.

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -208,19 +242,22 @@ public bool MatchLowercase(string text, int offset = 0)
...
+        public readonly bool MatchLowercase(string text, int end, int offset)
         {
             var index = Start + offset;
-            int i = 0;
-            for (; index <= end && i < text.Length; i++, index++)
+
+            if (end - index + 1 < text.Length)
+                return false;
+
+            string sliceText = Text;
```
xoofx: All the work you have done in this PR to move field access to single local variable is indeed helping a lot the codegen, amazing job

xoofx:
Amazing work on perf @MihaZupan with this PR! 👍

MihaZupan:
I was a bit sad there's no noticable overall effect from https://github.com/lunet-io/markdig/issues/363, so I only included the simple optimizations on IsDigit, IsAlpha

PR #386, opened 2020-01-12
Emojis and smileys customization

mlaily:
Hello,

I found out while updating my nuget packages that emojis and smileys customization was not possible anymore (#308).

This PR is an attempt to bring this feature back, while keeping the new performance optimizations in place.

Please note that I'm not much used to caring about allocations, so it's possible I introduced useless ones. Feel free to point them to me so I can fix them.

---

To summarize the changes:
- I introduced a new immutable `EmojiMapping` class, wrapping the creation of the compact prefix trees.
- `EmojiMapping` can be constructed from custom dictionaries (it uses the default emojis and smileys otherwise).
- `EmojiMapping` exposes two static public methods returning the default emojis and smileys dictionaries, so that they can be used as a base to create a custom `EmojiMapping` instance.
- `EmojiParser` takes an `EmojiMapping` instance, and uses it directly to parse emojis and smileys.

The compact prefix trees creation is mostly the same as before. I only added new safety checks, since the input can be provided externally.

I also added some unit tests ensuring the basic use-cases work as expected.


MihaZupan:
If we are adding support to customize the emoji set, I don't think having an EnableSmiley flag on EmojiParser makes sense anymore. I would still keep it in the UseEmojiAndSmiley and select the EmojiMapping to pass to the parser there.

EmojiMapping currently represents two possible mappings (smiley or not), I would prefer it only represents one and we instead have two default properties.

src/Markdig/Extensions/Emoji/EmojiParser.cs
```diff
@@ -76,1726 +70,5 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
 
             return true;
         }
-
-        #region Emojis and Smileys
-        static EmojiParser()
```
MihaZupan: It's good to see this static ctor go away

src/Markdig/Extensions/Emoji/EmojiParser.cs
```diff
@@ -40,8 +34,8 @@ public EmojiParser(bool enableSmiley = true)
 
         public override void Initialize()
```
MihaZupan: Initialize could be removed and the logic moved to the ctor

src/Markdig/Extensions/Emoji/EmojiMapping.cs
```diff
@@ -0,0 +1,1784 @@
...
+    {
+        /// <summary>
+        /// The default emojis and smileys mapping.
+        /// </summary>
+        public static readonly EmojiMapping DefaultEmojiMapping = new EmojiMapping();
+
+        internal CompactPrefixTree<string> EmojiPrefixTree { get; }
+        internal CompactPrefixTree<string> EmojiSmileyPrefixTree { get; }
+        internal char[] EmojiOpeningCharacters { get; }
+        internal char[] EmojiSmileyOpeningCharacters { get; }
```
MihaZupan: Nit: readonly

src/Markdig/Extensions/Emoji/EmojiMapping.cs
```diff
@@ -0,0 +1,1784 @@
...
+
+        #region Emojis and Smileys
+
+        /// <summary>
+        /// Returns a new instance of the default emoji to unicode dictionary.
+        /// It can be used to create a customized <see cref="EmojiMapping"/>.
+        /// </summary>
+        public static IDictionary<string, string> GetDefaultEmojiToUnicode()
```
MihaZupan: This can be Dictionary instead of the interface

src/Markdig/Extensions/Emoji/EmojiMapping.cs
```diff
@@ -0,0 +1,1784 @@
...
+                {":tanzania:","🇹🇿"},
+                {":ukraine:","🇺🇦"},
+                {":uganda:","🇺🇬"},
+                {":us_outlying_islands:","🇺🇲"},
+                {":united_nations:","🇺🇳"},
+                {":uruguay:","🇺🇾"},
+                {":uzbekistan:","🇺🇿"},
+                {":vatican_city:","🇻🇦"},
+                {":st_vincent_grenadines:","🇻🇨"},
+
```
MihaZupan: The ctor accepting a dictionary can be used directly here, otherwise the parameters should be `emojiToUnicode.Count, emojiToUnicode.Count * 2, emojiToUnicode.Count * 2` to save a few allocations on resizes.

MihaZupan: It's not obvious, but the `PrefixTree.Add()` will throw for null or empty keys already

MihaZupan:
Thank you for this, in general LGTM other than some implementation details mentioned above.

src/Markdig/Extensions/Emoji/EmojiMapping.cs
```diff
@@ -0,0 +1,1784 @@
...
+        #region Emojis and Smileys
+
+        /// <summary>
+        /// Returns a new instance of the default emoji to unicode dictionary.
+        /// It can be used to create a customized <see cref="EmojiMapping"/>.
+        /// </summary>
+        public static IDictionary<string, string> GetDefaultEmojiToUnicode()
+        {
+            return new Dictionary<string, string>()
```
MihaZupan: We could hardcode the capacity here (same for smileys below)

mlaily:
Thanks for the review.

> If we are adding support to customize the emoji set, I don't think having an EnableSmiley flag on EmojiParser makes sense anymore. I would still keep it in the UseEmojiAndSmiley and select the EmojiMapping to pass to the parser there.

My understanding is that it's currently possible to change the value of `EmojiParser.EnableSmiley` dynamically, and it should work. What you suggest would make this impossible, right? (Just checking. I'm not against it)

Also, what would `UseEmojiAndSmiley` look like? one override that takes a `bool enableSmiley` and uses the default mappings, and another one that takes just an `EmojiMapping` that can be customized?

> EmojiMapping currently represents two possible mappings (smiley or not), I would prefer it only represents one and we instead have two default properties.

So an `EmojiMapping` would instead be an optimized representation of a dictionary, containing only a compact prefix tree and an array of opening characters?
Maybe its name should change then?

What would be the two default properties ? One just for the emojis, and the second for both the emojis and the smileys?

src/Markdig/Extensions/Emoji/EmojiMapping.cs
mlaily: You mean a `readonly` field instead of a read-only property? I can't just add `readonly` here.

Why do you want that? There is a warning against it https://docs.microsoft.com/en-us/visualstudio/code-quality/ca2105

src/Markdig/Extensions/Emoji/EmojiMapping.cs
mlaily: Isn't it better to use the interface when possible and where it makes sense?

I used the interface because the dictionary can still be used (elements added and modified) with `IDictionary`, and it matches the type in the ctor of `EmojiMapping`.

src/Markdig/Extensions/Emoji/EmojiMapping.cs
mlaily: I didn't use the ctor here because I'm looping anyway to detect first characters. I thought that it would be more clear to also add at the same time, in the loop. What do you think?

(I will add the other parameters)

mlaily: Oh ok. Do you prefer I remove this line then? It's not totally useless as it outputs a nicer message, with the name of the parameter in question.

src/Markdig/Extensions/Emoji/EmojiMapping.cs
MihaZupan: Mb, missed that you made it a property

src/Markdig/Extensions/Emoji/EmojiMapping.cs
MihaZupan: Up to personal preference, if you find it more readable I'm fine with that as long as the parameters are updated (the ctor will choose the same numbers and foreach over it the same way)

MihaZupan: We can keep it, just move it up so that it will be hit first

MihaZupan:
> change the value of EmojiParser.EnableSmiley dynamically

No. It can technically be changed between the ctor and first use, but that was to allow for customization before. Now we don't need to wait for initialize, that work can be done in the ctor.

In general, parsers **should** be immutable and thread-safe, to allow for Pipeline reuse to be safe.

> Also, what would UseEmojiAndSmiley look like? one override that takes a bool enableSmiley and uses the default mappings, and another one that takes just an EmojiMapping that can be customized?

That was the idea. The one accepting bool would call the one accepting EmojiMapping directly.


> Maybe its name should change then?

I think it's still a good name, no hard preference tho

> What would be the two default properties ? One just for the emojis, and the second for both the emojis and the smileys?

Yes

src/Markdig/Extensions/Emoji/EmojiMapping.cs
mlaily: Did you miss my comment? Any thought?

MihaZupan: Ah yeah I missed it. I suppose it doesn't really make a difference in this case.

mlaily:
I'm working on the changes, but I also have another question: I don't think the terminology used in the extension is correct.

Since I'm introducing new properties, I would like to use this as an opportunity to use a better terminology.

Specifically, I don't think "emoji" is used properly. I think "emoji" should be used to talk about the unicode representation.
To replace the current meaning in the extension, I suggest using "shortcode" instead. This seem to be the widely used term to describe things like `:confused:`. https://emojipedia.org/shortcodes/

I'm also not a big fan of "smiley" to describe a face made of text, but I don't have anything better to suggest...

What do you think?

I would make these changes in a separate commit at the end...

MihaZupan:
The naming change LGTM, I can't really think of a different name for smiley.

mlaily:
That's alright, thank you :)

MihaZupan:
Well that's an awful git merge diff, sorry. Looks like the web editor replaced all \n with \r\n

src/Markdig/Extensions/Emoji/EmojiMapping.cs
xoofx: @mlaily could you add the correct capacities, I will merge the PR after that, thanks!

mlaily: Oh, sorry, I missed that.

mlaily:
I rebased my branch on top of the markdig master. The only conflict I had was the changelog.

If for whatever reason you don't like that, tell me and I'll revert the rebase, and do a merge into my branch instead.

xoofx:
Thanks for that!

mlaily:
Thank **you** for your wonderful work on open source projects!

xoofx:
> Thank you for your wonderful work on open source projects!

I forgot to thank also @MihaZupan who is helping a lot to keep this project updated. Thank you for again reviewing this PR @MihaZupan !

PR #388, opened 2020-01-21
Optimized LineReader using string.IndexOfAny

KrisVandermotten:
The `LineReader` struct is responsible for parsing source text into lines.

Given that `String.IndexOfAny` [is now vectorized](https://source.dot.net/#System.Private.CoreLib/String.Searching.cs,cf166f50515b56ae), it makes sense to use it instead of the handwritten, unoptimized loop in `LineReader.ReadLine()`.

### DIFF
--- a/src/Markdig/Helpers/LineReader.cs
+++ b/src/Markdig/Helpers/LineReader.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Alexandre Mutel. All rights reserved.
+// Copyright (c) Alexandre Mutel. All rights reserved.
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
 
@@ -13,6 +13,8 @@ namespace Markdig.Helpers
     /// </summary>
     public struct LineReader
     {
+        private static readonly char[] lineDelimiters = new char[] { '\r', '\n' };
+
         private readonly string text;
 
         /// <summary>
@@ -38,34 +40,31 @@ namespace Markdig.Helpers
         /// <returns>A new line or null if the end of <see cref="TextReader"/> has been reached</returns>
         public StringSlice? ReadLine()
         {
-            if (SourcePosition >= text.Length)
+            int startPosition = SourcePosition;
+
+            if (startPosition >= text.Length)
             {
                 return null;
             }
 
-            var startPosition = SourcePosition;
-            var position = SourcePosition;
-            var slice = new StringSlice(text, startPosition, startPosition);
-            for (;position < text.Length; position++)
+            int position = text.IndexOfAny(lineDelimiters, startPosition);
+
+            if (position < 0)
             {
-                var c = text[position];
-                if (c == '\r' || c == '\n')
-                {
-                    slice.End = position - 1;
-                    if (c == '\r' && position + 1 < text.Length && text[position + 1] == '\n')
-                    {
-                        position++;
-                    }
-                    position++;
-                    SourcePosition = position;
-                    return slice;
-                }
+                SourcePosition = text.Length;
+                return new StringSlice(text, startPosition, text.Length - 1);
             }
-
-            slice.End = position - 1;
-            SourcePosition = position;
-
-            return slice;
+            else
+            {
+                var slice = new StringSlice(text, startPosition, position - 1);
+                if (text[position] == '\r' && position + 1 < text.Length && text[position + 1] == '\n')
+                {
+                    position++;
+                }
+                position++;
+                SourcePosition = position;
+                return slice;
+            }
         }
-   }
+    }
 }


MihaZupan:
Thanks for pointing this one out. As the optimization is only applied on Core, I think we can use an ifdef to only use it there.

KrisVandermotten:
`String.IndexOfAny` is implemented in native code on .NET Framework. I would not assume that is is slower than the old loop without benchmarking it (I haven't).

xoofx:
@KrisVandermotten there is a benchmark project in the src folder. You can comment only to run Markdig and double check on both .NET Core and .NET Standard how much this is actually bringing.

xoofx:
You will need to rebase on master as I have change the CI.

KrisVandermotten:
I benchmarked various conbinations, using the following tests:

    public class Program
    {
        private string spec = File.ReadAllText("spec.md");

        [Benchmark]
        public void TestSpec()
        {
            var lineReader = new LineReader(spec);

            while (lineReader.ReadLine().HasValue)
            {
            }
        }

        private string longLine = new string('a', 4096);
        
        [Benchmark]
        public void TestLongLine()
        {
            var lineReader = new LineReader(longLine);

            while (lineReader.ReadLine().HasValue)
            {
            }
        }

        static void Main(string[] args)
        {
        }
    }

# Results

These are the results on .NET Framework:

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
  [Host]     : .NET Framework 4.8 (4.8.4075.0), X64 RyuJIT
  DefaultJob : .NET Framework 4.8 (4.8.4075.0), X64 RyuJIT

Using the original code:

|       Method |       Mean |     Error |    StdDev |
|------------- |-----------:|----------:|----------:|
|     TestSpec | 325.901 us | 1.0245 us | 0.9082 us |
| TestLongLine |   2.144 us | 0.0126 us | 0.0117 us |

Using `IndexOfAny`:

|       Method |       Mean |     Error |    StdDev |
|------------- |-----------:|----------:|----------:|
|     TestSpec | 528.317 us | 5.7727 us | 5.1173 us |
| TestLongLine |   3.978 us | 0.0231 us | 0.0193 us |

These are the results on .NET Core 3.1.1:

BenchmarkDotNet=v0.12.0, OS=Windows 10.0.18363
Intel Core i7-7700 CPU 3.60GHz (Kaby Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.1.101
  [Host]     : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT
  DefaultJob : .NET Core 3.1.1 (CoreCLR 4.700.19.60701, CoreFX 4.700.19.60801), X64 RyuJIT

Using the original code:

|       Method |       Mean |     Error |    StdDev |
|------------- |-----------:|----------:|----------:|
|     TestSpec | 330.042 us | 0.9974 us | 0.8329 us |
| TestLongLine |   2.138 us | 0.0113 us | 0.0106 us |

Using `IndexOfAny`:

|       Method |         Mean |        Error |       StdDev |
|------------- |-------------:|-------------:|-------------:|
|     TestSpec | 377,382.5 ns | 14,748.87 ns | 13,796.10 ns |
| TestLongLine |     263.2 ns |      1.24 ns |      1.16 ns |

# Conclusions

On .NET Framework, using `IndexOfAny` is always slower than the original code.

Even on .NET Core, the original code is usually faster than using `IndexOfAny`. Only when the source text contains long lines, the vectorization pays of and parsing can be about 8 times faster (due to the 8-fold parallellism - on my machine).

This is of course just for the parsing of the source text into separate lines. The test did not include parsing those lines or rendering output.

For those reasons, I recommend not making this change, not even for .NET Core.

What do you think?

xoofx:
> What do you think?

Thanks for the benchmark!

I'm not entirely surprised. I remember that I did benchmark `IndexOfAny` and that's why I end-up using a manual version of it... (and mentioned this  in the [blog post](https://xoofx.com/blog/2016/06/13/implementing-a-markdown-processor-for-dotnet/)) but was not sure now about the results.

We can wait for @MihaZupan to find out if there is no possible improvements but otherwise, we will close it.

MihaZupan:
Some interesting findings

First, I tried to help out the Jit with range check elimination by rewriting the method to
```c#
public StringSlice? ReadLine()
{
    string text = _text;
    int i = SourcePosition;

    for (; i < text.Length; i++)
    {
        char c = text[i];
        if (c == '\r' || c == '\n')
        {
            var slice = new StringSlice(text, SourcePosition, i - 1);

            if (c == '\r' && (uint)(i + 1) < (uint)text.Length && text[i + 1] == '\n')
                i++;

            SourcePosition = i + 1;
            return slice;
        }
    }

    if (SourcePosition >= text.Length)
        return null;

    var remainder = new StringSlice(text, SourcePosition, i - 1);
    SourcePosition = int.MaxValue;
    return remainder;
}
```

This gave me about a ~4% boost on the LineReader microbenchmark, running over the CommonMark spec file.

Testing with an IndexOfAny implementation, that pulled in an extra ~1.5% for CommonMark and 550% for a best-case scenario of only very long lines. But it does require an ifdef to only apply change in NetCore.

Then for some reason I tried changing the method signature from StringSlice? to StringSlice, and returning default instead of null. That change alone brought a **25% improvement**!

Then, instead of using `int i = SourcePosition`, extracting SourcePosition to a local copy and then only using `int i` in the for loop yields an extra 13.5% improvement.

With that, I recommend ignoring IndexOfAny as it doesn't really bring an improvement to real-world markdown files (and to be fair, the cost of LineReader is low compared to the rest of parsing). This way we get to keep a common implementation across targets and we don't even need to touch unsafe code.

Final implementation for reference, taking 61% of the time in the microbenchmark compared to the current impl
```c#
public StringSlice ReadLine()
{
    string text = _text;
    int sourcePosition = SourcePosition;

    for (int i = sourcePosition; i < text.Length; i++)
    {
        char c = text[i];
        if (c == '\r' || c == '\n')
        {
            var slice = new StringSlice(text, sourcePosition, i - 1);

            if (c == '\r' && (uint)(i + 1) < (uint)text.Length && text[i + 1] == '\n')
                i++;

            SourcePosition = i + 1;
            return slice;
        }
    }

    if (sourcePosition >= text.Length)
        return default;

    var remainder = new StringSlice(text, sourcePosition, text.Length - 1);
    SourcePosition = int.MaxValue;
    return remainder;
}
```
This is a breaking change in the LineReader api, check for end-of-file is `result.Text is null` instead of `result == null`.

xoofx:
> This is a breaking change in the LineReader api, check for end-of-file is result.Text is null instead of result == null.

Great findings, you can break the API, Markdig is a 0.x API, so we can do whatever is necessary to level-up during this period.

PR #389, opened 2020-01-22
Improve code coverage

MihaZupan:
Add some tests and fix a few details that those tests uncovered.

I think I can push coverage to ~98% with reasonable tests like this, the rest would be catching every edge-case handling logic.


xoofx:
Amazing, thanks for that! 👍

PR #391, opened 2020-01-26
Remove UAP specific code

MihaZupan:


xoofx:
Ha, thanks for that, completely forgot about removing these when removing UAP from csproj.

PR #393, opened 2020-01-30
Optimize LineReader.ReadLine

MihaZupan:
Closes #388

Discussion for reference: https://github.com/lunet-io/markdig/pull/388#issuecomment-580429867


xoofx:
Thanks!

PR #399, opened 2020-02-15
Some optimizations in StringLineGroup

KrisVandermotten:
This PR contributes a few improvements to `StringLineGroup`.

#### `RemoveAt(int)`

A memory leak was fixed and code duplication reduced.

#### `ToSlice(List<LineOffset>)`

First, the case where `Count == 1` has been moved before the case where `Count == 0`, simply because it is so much more common. Indeed, out of 1023 unit tests that call this method, 872 execute the `Count == 1` branch, but none of them execute the `Count == 0` branch.

Secondly, the loop was restructured to simplify it and reduce code duplication. Using `ref var line = ref Lines[i];`, the `Lines` array is indexed only once per iteration.

#### `Iterator.NextChar()`

`_offset++;` has been moved into the `if` branch, so that the `_offset--;` in the else branch can be removed.

### DIFF
--- a/src/Markdig/Helpers/StringLineGroup.cs
+++ b/src/Markdig/Helpers/StringLineGroup.cs
@@ -69,16 +69,13 @@ namespace Markdig.Helpers
         /// <param name="index">The index.</param>
         public void RemoveAt(int index)
         {
-            if (Count - 1 == index)
-            {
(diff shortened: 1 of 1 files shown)


xoofx:
Thanks!

PR #402, opened 2020-03-01
Fixed a bug in HeadingRenderer

KrisVandermotten:
The rendered HTML for Heading level 7 and above was incorrect.

This PR fixes that.

In addition, a minor optimization removes the range check over the array bounds.


MihaZupan:
> In addition, a minor optimization removes the range check over the array bounds.

Sadly I don't think it does, unless you save HeadingTexts to a local first.

KrisVandermotten:
> Sadly I don't think it does, unless you save HeadingTexts to a local first.

Actually I think it will, at least on some runtimes, because the field is `static readonly`, though I must admit I didn't verify. Eliminating the range check wasn't the point of the PR.

And even if the range check doesn't get eliminated, at least two comparisons have been replaced by one. 🙂

PR #409, opened 2020-03-18
Don't copy BitVector128 instances, to work around UWP bug

### LINKED ISSUE #408: BitVector128 not copying consistently
Here's an odd one.  I have a .NET Framework 4.6.2 test program using markdig 0.18.3 and it worked fine.  I have a UWP app (target 18362) using the same version of markdig and it fails to recognize auto links and other things.

I have traced at least one problem down to line 79 of CharacterMap.cs:

`this.isOpeningCharacter = isOpeningCharacter;`

In the .NET program, this line copies all 4 uint values of the vector.  In the UWP version, only the first uint is copied.  That is, the x86 code generated for the copy is different.

BitVector128 was introduced in 0.18.2.  I have reverted to 0.18.1 and the problem has disappeared.

KrisVandermotten:
Hopefully, this fixes #408. 

Important: I did not test this in an actual UWP program.

### DIFF
--- a/src/Markdig/Helpers/CharacterMap.cs
+++ b/src/Markdig/Helpers/CharacterMap.cs
@@ -51,7 +51,7 @@ namespace Markdig.Helpers
             Array.Sort(OpeningCharacters);
 
             asciiMap = new T[maxChar + 1];
-            var isOpeningCharacter = new BitVector128();
+            isOpeningCharacter = new BitVector128();
 
             foreach (var state in maps)
             {
@@ -75,8 +75,6 @@ namespace Markdig.Helpers
                     }
                 }
             }
-
-            this.isOpeningCharacter = isOpeningCharacter;
         }
 
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/Markdig/Helpers/CharacterMap.cs
```diff
@@ -127,7 +123,7 @@ public int IndexOfOpeningCharacter(string text, int start, int end)
                         for (int i = start; i <= end; i++)
                         {
                             var c = pText[i];
-                            if (c < 128 && openingChars[c])
+                            if (c < 128 && isOpeningCharacter[c])
```
MihaZupan: I don't like that we're now fetching a field every time. We should try to see if the first half of this PR (ctor change) fixes the UWP issue

MihaZupan:
Closing this as it's a [UWP issue](https://github.com/lunet-io/markdig/issues/408#issuecomment-600883345), we shouldn't try to work around it until we understand it.

Issue #296, opened 2019-01-26
Abbreviation matching


MihaZupan:
I've been looking at the `AbbreviationParser` and came accross this [discrepancy](https://talk.commonmark.org/t/abbreviations-longest-shortest-valid-match/3060) between different versions.

As a side-note, markdig currently doesn't match at all in cases like [this](https://babelmark.github.io/?text=*%5BFoo%5D%3A+foo%0A*%5BFoo+Bar%5D%3A+foobar%0A%0AFoo+B) due to how the `TextMatchHelper `works internally.

I'd be happy to fix this, just wanted to hear your thoughts on the commonmark issue first.

xoofx:
You can go ahead with what you think is more relevant/safe/consistent. The `TextMatchHelper` is one of the helper classes that would definitely need to be revisited 👍

MihaZupan:
I've written a version of TextMatchHelper as a [CompactPrefixTree](https://github.com/MihaZupan/SharpCollections).
Would you prefer I port it to markdig and add it as a helper class or add a NuGet reference (cross targets all platforms Markdig does)?

xoofx:
> Would you prefer I port it to markdig and add it as a helper class or add a NuGet reference (cross targets all platforms Markdig does)?

Yeah, port it, I would like to avoid any dependencies.

PR #416, opened 2020-04-07
Random improvements

MihaZupan:
- Dropping support for `net35`, `net40`. Now multi-targeting `netstandard2.0`, `netcoreapp2.1,` `netcoreapp3.1`. 🎉
- Removed helpers such as `ArrayHelper`, `MethodImplOptionPortable`, `ExcludeFromCodeCoverage` that were needed for framework target compat.
- Unsealed `MarkdownParserContext` (@patriksvensson I don't suppose you somehow rely on it being sealed? I figure having it unsealed adds more value)
- Removed some allocations (b090f24, 84ec832, bdfaa27, 30d0924), 2% overall.
- Improved codegen for [some StringSlice helpers](https://github.com/lunet-io/markdig/commit/f0830b3237b51e4297bc93ce0954ce80eb24cf15).
- Used `ThrowHelper` [everywhere](https://github.com/lunet-io/markdig/commit/a99e65004c1faf7304472829a863f1b55dfe6e38), hopefully more things can inline now.
- Further optimized `CharacterMap` in 256b4d3. @craigwi while not the reason I did it, this may also workaround the UWP issue.

Overall ~6% parsing throughput increase. Allocation-wise we are now very close to noise-free. The vast majority of allocations are now the AST itself and some strings.


KrisVandermotten:
You changed `BitVector128`, a 16 byte struct, to `BoolVector128`, a 512 byte struct. Are you sure you want to do that?

MihaZupan:
The allocation overhead of 128 bytes is negligible as we don't construct character maps often. Calls to CharacterMap methods don't copy the struct either so I don't see a downside.

craigwi:
Lots of interesting changes.  My recommendation is to split the PR into about a half of a dozen smaller PR, each of which focuses on a specific change to make it easier to review.

MihaZupan:
I tried to keep changes isolated between commits and I don't think they should be too contentious (except perhaps a99e650). Is there any particular change you would find easier to review if I isolate it into a separate PR?

KrisVandermotten:
> The allocation overhead of 128 bytes is negligible as we don't construct character maps often. Calls to CharacterMap methods don't copy the struct either so I don't see a downside.

Well, the struct no longer fits in a single cache line.

Note that going from a 128 byte array to a 16 byte struct resulted in a 3% throughput improvement for the parsing.

>  Is there any particular change you would find easier to review if I isolate it into a separate PR?

This might be such a change yes. It would be interesting to see before and after benchmark results.

MihaZupan:
> This might be such a change yes. It would be interesting to see before and after benchmark results.

You got it, I'll move it out. The biggest chunk of throughput improvement from this PR comes from character map changes.

MihaZupan:
> Should we not ensure that the array cannot grow without bounds?

We should, looks like I over-simplified that part.

patriksvensson:
@MihaZupan Just an old habit of sealing everything until there's a reason for it to be unsealed.

MihaZupan:
@KrisVandermotten I've extracted it to https://github.com/MihaZupan/markdig/pull/1, you can see the difference in the generated asm.

KrisVandermotten:
@MihaZupan I'm sorry I did not make myself clear. I'm curious about the effect is of only the change from the 16 byte struct to 512 byte struct. I have no doubt that some of the other changes you did there are beneficial.

FYI, I may find some time later this week to take a look myself as well.

MihaZupan:
> I'm curious about the effect is of only the change from the 16 byte struct to 512 byte struct

Ah I see. The struct is 128 bytes, not 512. I can run a benchmark on just this change but I would expect the bool lookup to still be (not much) faster.

> FYI, I may find some time later this week to take a look myself as well.

Glad to hear it!

KrisVandermotten:
> The struct is 128 bytes, not 512.

I stand corrected. Somehow I was convinced that `sizeof(bool) == 4`. It goes without saying that the difference is significant.

MihaZupan:
cc @Sergio0694

Sergio0694:
@MihaZupan **NICE!** 🍻🚀

A couple questions:
1) Why don't you also add the `netstandard.2.1` target while you're at it? Have you checked if there's some overlap with the .NET Core 3.1 APIs you're using and .NET Standard 2.1, so that those new features could also be available in the portable version?
2) Regarding the `ThrowHelpers`, not sure if taking an external dependency is out of the questions, but if isn't, you might be interested in referencing `Microsoft.Toolkit` when thre `6.1` update is out, and checking out the new `Guard` APIs ([PR here](https://github.com/windows-toolkit/WindowsCommunityToolkit/pull/3131)), using those could simplify a lot of these conditional checks and remove some more code from `markdig`. They're already pretty optimized with respect to the codegen paths. Let me know if that helps! 😄

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -312,16 +304,15 @@ public bool SkipSpacesToEndOfLineOrEndOfDocument()
...
-                {
-                    break;
-                }
-            }
-            return IsEmpty;
+            string text = Text;
+            int end = End;
+            int i = Start;
+
+            while (i <= end && (uint)i < (uint)text.Length && text[i].IsWhitespace())
```
Sergio0694, line 311: I might be missing something, but why not leverage the `ReadOnlySpan<T>.Trim` extension (from `MemoryExtensions`) here instead, which is also automatically vectorized? Same for the other trim methods from this type.

MihaZupan:
> Why don't you also add the netstandard.2.1 target while you're at it? Have you checked if there's some overlap with the .NET Core 3.1 APIs you're using and .NET Standard 2.1, so that those new features could also be available in the portable version?

We can target netstandard2.1 as well, I added core 3.1 for the `string.GetPinnableReference` API, I don't know of an alternative to get `ref char` from a string (without going through Span / pinning).

> Regarding the ThrowHelpers, not sure if taking an external dependency is out of the questions, but if isn't, you might be interested in referencing Microsoft.Toolkit

Considering the simplicity of those methods, I wouldn't add a dependency. See https://github.com/lunet-io/markdig/issues/296#issuecomment-461005586.

src/Markdig/Helpers/StringSlice.cs
MihaZupan: I would expect the 99% case here to be 0 - 2 spaces, so creating the span would be more overhead.

I have played around with other unsafe helpers (that could be used in `Match` for example), that would take `ref char` to avoid the cost of a span creation, but that's for a different issue/pr.

Sergio0694:
> 

If you don't mind doing dirty tricks, there's this 🙈

https://github.com/Sergio0694/WindowsCommunityToolkit/blob/feature/high-perf-apis/Microsoft.Toolkit.HighPerformance/Extensions/StringExtensions.cs#L26-L31

This also happens to be faster than `GetPinnableReference`, since it can skip the bounds check, though technically it is only guaranteed to work on CoreCLR-like runtimes. But, since you're using that inside the `#if NETCOREAPP3_1` directive anyway, it should be fine. I understand if you didn't want to have stuff like this in the repo for maintanability though 😅

> Considering the simplicity of those methods, I wouldn't add a dependency. See #296 (comment).

Yup, figured as much, makes perfect sense! 👍

MihaZupan:
string's `GetPinnableReference` doesn't do any validation afaik ([sharplab](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAOgCUBXAOwwEt8YaBhCfAB1YAbGFADKIgG6swMXAG4AsAChlxAMyUqSSgCZyXcgG9l5U+RNn1m7bABm5MAAtsUcgHEYGOjFsiYTGQAKDEdWXE0KDBgEDABKCzMzAF4APnI7cgBVJlxsXxoAQVxvW0DWJnIomJoPDAAFcqZsYGESvwCYQNjYxSUAXyA==)). 

> If you don't mind doing dirty tricks, there's this 🙈

👀 Thanks for sharing this. While I don't mind unsafe, I would rather have a preprocessor directive here in this case while we support targets without the API 😄

Sergio0694:
👀 Well thank you for sharing that too!

I was sure `string.GetPinnableReference()` was doing a check as well, but apparently I was wrong and that only happens for `Span<T>.GetPinnableReference()` (see [here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4aNYgGYm5JE1IMhDAN40Glhhas69B2ADMGYABbYoDAOIwMbGI4yMFwqABQYrry4eowYMAgYAJQ2VlYAvAB8DE4MAKpcuNiBLACCuP6OobxcDHEJLD4YAArVXNjA0hVBITChiYka1KkpTLrMDgHZMNgAJhBckgCeLu6ejV2wPeGR0f6zAPILi7Ji2FwAPG4eWbinXMlDqelZObdnDb4tXG0dMBvBYQGNAAvkA=)). That's a bit weird, I wonder why do those APIs act so differently (assuming it's not just sharplab.io acting up). I'll need to go find the source in CoreCLR, hopefully there will be some comments there that will shed some light on this! 😄

Well what do you know! Here's the source from CoreCLR:

https://github.com/dotnet/runtime/blob/master/src/libraries/System.Private.CoreLib/src/System/String.cs#L490-L495

This is great, thanks again! TIL 🍻

Sergio0694:
Small note on that `string.GetPinnableReference()`, while it's great to have it built-in, looks like it unfortunately suffers from a very similar codegen issue to https://github.com/dotnet/runtime/issues/12880.

In particular, we get this in x64:

```csharp
public static ref char GetReference(this string text)
{
    return ref Unsafe.AsRef(in text.GetPinnableReference());   
}
```

```asm
C.GetReference(System.String)
    L0000: mov eax, [rcx]
    L0002: lea rax, [rcx+0xc]
    L0006: ret
```

That first `mov eax, [rcx]` looks unnecessary, since the value in `eax` is immediately overwritten and never actually used. Also I'm not even sure what that was supposed to be in the first place, since it's apparently reading a 32bit value (in `eax`), from the address of... The method table pointer? The first `char` is located at `[rcx + 0xc]` since that's 8 bytes for the method table pointer (x64) + 4 bytes for the `int` representing the length of the `string`. Might be a codegen leftover for the bounds check 🤔

Anyway, hopefully it gets fixed in a future RyuJit update, thought it might be interesting to share 😄

xoofx:
>     `L0000: mov eax, [rcx]` 

I should be a check for `NullReferenceException` so that access to `text` is still safe.

Sergio0694:
@xoofx Oh I see, thanks! Is that because if the address was null, trying to dereference `[rcx+0xc]` would have resulted in an access violation instead? As in, since that address would've been invalid but not exactlu `null`, so I'm not sure if the runtime is able to handle that the same? 🤔

xoofx:
> @xoofx Oh I see, thanks! Is that because if the address was null, trying to dereference [rcx+0xc] would have resulted in an access violation instead? As in, since that address would've been invalid but not exactlu null, so I'm not sure if the runtime is able to handle that the same?

```
    L0002: lea rax, [rcx+0xc]
```
This instruction doesn't dereference anything but just calculate an offset. That means that if the first instruction `mov eax, [rcx]` was not emitted, you would return `rax = 12` without any error in this function. So the problem is that you really want to know that `text.GetPinnableReference()` failed on a `NullReferenceException` not much later when accessing the pointer returned by `GetReference`

An access to the memory behind 0xc or 0x0 should generate a `NullReferenceException`, usually because it is in the first range of a 4K page fault that are mapped to generate such exception.

Sergio0694:
Ah, gotcha, that makes perfect sense. Also, good to know that memory accesses in that initial range would still properly trigger a `NullReferenceException` instead of a nasty AV, that's nice 😄

I guess if one really wanted to also skip that `null` check there, the solution would just be to either use the trick with the mapping type ([sample here](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACY8gOgCUBXAOwwEt8YLAMIR8AB14AbGFADKMgG68wMXAG4a9Jq049+ggJI8ZEMfKhKV6mpoDM2pE1IMhNAN40GXpveaPYAGYMYAAW2FAMAOIwGGwwATIwXCoAFBghvLjajBgwCBgAlJ7eHtTe5UwA7AyBDACqXLjYCSwAgrhxASm8XAy5+SzRGAAKPVzYwNKdickwKQUFat7FXgC+KwwbxL7k/vE1MNgAJhBckgCewWERACLYXADmJhy4Q9Ows2kZWcw5eYUbUoVbwKcLZO4YbAMAC89UazUE7QAPGxsAB3WQYKA9B4Q7AAPjS/0WNjKwKqByCvzxLDxGjJa1J5QA2pioBwwBgADLYc4QDgYFI8vkCgDSPSOLAAoggxJJlLxCgBdDZibGg3IMXCHaRHJwMVEYrE4vGAjYsgBivBgkiOAHkAgFtYKUAUVQyKtsrmC6Rt1tRVkA=)), or to just wait for .NET 5 (maybe?) to add some `GetRawDataReference` API or equivalent for the `string` type. Something like `GetRawData` (https://github.com/dotnet/runtime/issues/28001), which for now is only for arrays as far as I know ([`GetArrayDataReference`](https://github.com/dotnet/runtime/blob/8cc28d7ee63ae6f07dd9a37b1b8a935be565076e/src/coreclr/src/System.Private.CoreLib/src/System/Runtime/InteropServices/MemoryMarshal.CoreCLR.cs#L24)).

Of course these are arguably just very minor details, but mostly I was just interested in learning more about this in general. Thanks for taking the time to chime in, really appreciate it! 😊

KrisVandermotten:
I've been benchmarking the difference between the 128 bool array and the 16 byte array.

It seems that the bool array is slightly faster (at least on my computer, .NET Core 3.1.3) on this benchmark:

    public class Program
    {
        private readonly string text = File.ReadAllText("spec.md");

        [Benchmark]
        public void TestMarkdig()
        {
            Markdown.Parse(text);
        }

        private static void Main(string[] args)
        {
        }
    }

KrisVandermotten:
LGTM

xoofx:
Could you re-merge back master to your branch @MihaZupan?, the CI was broken and I had to fix it on master.  You might have also to rerun the specs after merging. Thanks!

MihaZupan:
Thanks for fixing up the CI! I rebased and also added a commit to cross-target standard2.1 as Sergio suggested

src/Markdig/Helpers/ThrowHelper.cs
```diff
@@ -10,45 +10,46 @@ namespace Markdig.Helpers
     [ExcludeFromCodeCoverage]
     internal static class ThrowHelper
     {
+        public static void ArgumentNullException(string paramName) => throw new ArgumentNullException(paramName);
```
Sergio0694, line 13: I might be missing something, but shouldn't all these methods be marked as `[MethodImpl(MethodImplOptions.NoInlining)]`?

src/Markdig/Markdig.targets
```diff
@@ -6,7 +6,7 @@
     <NeutralLanguage>en-US</NeutralLanguage>
     <VersionPrefix>0.18.3</VersionPrefix>
     <Authors>Alexandre Mutel</Authors>
-    <TargetFrameworks>net35;net40;netstandard2.0;netcoreapp2.1</TargetFrameworks>
+    <TargetFrameworks>netstandard2.0;netstandard2.1;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
```
Sergio0694, line 9: Out of curiosity, what .NET Core 2.1 specific APIs are you using in the lib?
Wouldn't .NET Standard 2.0 be enough to replace that target too?

And as for .NET Core 3.1, I assume that's because you have some vectorized paths using SSE/AVX/etc.?

src/Markdig/Helpers/ThrowHelper.cs
MihaZupan: They could be, but the Jit will never find inlining methods that always throw profitable afaik

src/Markdig/Markdig.targets
MihaZupan: Some APIs like `TextWriter.Write(Span)` aren't in standard 2.0 so that's why I added core2.1 before.
Core 3.1 is used only for string.GetPinnableReference in CharacterMap rn

MihaZupan: Right now there's no manual hw intrinsics code in Markdig, but that can change now that we're refreshing build targets

Sergio0694: I see, targeting .NET Core 2.1 directly as a workaround to be able to use those `TextWriter.Write(Span<T>)` overloads makes sense, yeah. As for .NET Core 3.1 though, if it's just for that single `string.GetPinnableReference` call, wouldn't it make more sense to drop that target, just target .NET Standard 2.1, and pass `ReadOnlySpan<char>` values around (instead of `string`-s), and then just use `MemoryMarshal.GetReference(ReadOnlySpan<T>)` to get a mutable `ref` from there? This would also allow you to just split/tokenize over `ReadOnlySpan<T>` instead of allocating substrings, in case you're doing that too.

Just a thought! 😄

src/Markdig/Helpers/ThrowHelper.cs
Sergio0694: Ah I see, you're relying more on the JIT to do that for you. I guess I might just be overly cautious to always add that manually, just in case the code ends up running on some runtime with an older or less optimized JIT. But what you said makes perfect sense, yeah! 👍

MihaZupan: I can add them to document that that is the intent

Sergio0694: Yeah, it should be handled automatically in CoreCLR now (see https://github.com/dotnet/coreclr/pull/6103), but I guess it wouldn't hurt considering the lib targets .NET Standard, so it could be running on everything from .NET Framework to other runtimes as well 😊

src/Markdig/Markdig.targets
MihaZupan: StringSlice is the type plummed throughout Markdig to avoid substring allocations. It provides good utility like the NextChar method that you would otherwise have to replace with slices/more manual index bookkeeping.

I'm not too worried about having the extra target since I can see us adding more code relying on new APIs in the future

Sergio0694: I see, yeah that makes perfect sense. I was just curious, thank you for taking the time to provide more info, I appreciate it! 😊

xoofx:
Amazing job @MihaZupan !

PR #417, opened 2020-04-13
Cleanup StringBuilderCache

KrisVandermotten:
`StringBuilderCache` used to inherit from `DefaultObjectCache<StringBuilder>`, and instances were being passed around.

However, nobody was actually using those instances, with one exception: `CodeInlineParser`.

All other code uses `StringBuilderCache.Local()`. As it turns out, there is actually no reason why `CodeInlineParser` could not do the same.

This PR changes `StringBuilderCache` to be a `static class`, no longer inheriting from `DefaultObjectCache<StringBuilder>`. All references to instances are removed. `CodeInlineParser` was modified to use `StringBuilderCache.Local()` as well.

It goes without saying that this is a breaking change, as the class and most of the instance references were public.


xoofx:
Sorry, the CI on master was broken and it should be fixed now. Could you rebase/merge master to your branch please? Thanks!

xoofx:
Can you fix the conflicts?

xoofx:
Cool, thanks!

PR #425, opened 2020-04-19
Add IReadOnlyList interface to ContainerBlock to unify and simplify e…

lellid:
Problem: `ContainerBlock` implements the  `IList<Block>` interface, but this interface does not contain the `IReadOnlyList<Block>`

In order to simplify and unify the enumeration of child items, add the `IReadOnlyList<Block>` interface to  `ContainerBlock`. `ContainerBlock` already contains the neccessary implementations of the `IReadOnlyList<Block>` interface.


xoofx:
Thanks!

PR #426, opened 2020-04-21
git add --renormalize .

MihaZupan:
Hopefully the last diff with a ton of line-ending changes

With hidden whitespace changes: https://github.com/lunet-io/markdig/pull/426/files?diff=split&w=1


xoofx:
Han didn't know about this new `--renormalize` command, cool, thanks!

MihaZupan:
I just found out about it from https://github.com/microsoft/reverse-proxy/pull/35

PR #429, opened 2020-04-26
Added StringSlice.CountAndSkipChar(char)

KrisVandermotten:
This PR is mostly about extracting common logic into a reusable method.

Obviously, care was taken to implement the new method as efficient as possible.

### DIFF
--- a/src/Markdig/Parsers/FencedBlockParserBase.cs
+++ b/src/Markdig/Parsers/FencedBlockParserBase.cs
@@ -144,22 +144,10 @@ namespace Markdig.Parsers
                 return BlockState.None;
             }
 
-            var startPosition = processor.Start;
-
             // Match fenced char
-            int count = 0;
             var line = processor.Line;
-            char c = line.CurrentChar;
-            var matchChar = c;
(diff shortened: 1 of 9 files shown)


src/Markdig/Parsers/FencedBlockParserBase.cs
```diff
@@ -191,14 +179,7 @@ public override BlockState TryOpen(BlockProcessor processor)
             // Add the language as an attribute by default
             if (!string.IsNullOrEmpty(fenced.Info))
             {
-                if (string.IsNullOrEmpty(InfoPrefix))
```
xoofx, line 194: Don't recall and can't see it from the PR context, but InfoPrefix can never be empty?

MihaZupan: I think what he was getting at is that string concat with null or empty is basically a nop, so the if/else can be simplified.

KrisVandermotten: Indeed. string.Concat(string, string) tests this already.

PR #435, opened 2020-05-12
Fix media link classes not being added

RudeySH:
I noticed this when using both the Bootstrap extension and the Media Links extension. When Bootstraps adds `img-fluid`, the Media Link extension didn't add their classes as well. This PR should fix that.

### DIFF
--- a/src/Markdig/Extensions/MediaLinks/MediaLinkExtension.cs
+++ b/src/Markdig/Extensions/MediaLinks/MediaLinkExtension.cs
@@ -154,10 +154,12 @@ namespace Markdig.Extensions.MediaLinks
                 htmlAttributes.AddPropertyIfNotExist("height", Options.Height);
 
             if (!string.IsNullOrEmpty(Options.Class) || !string.IsNullOrEmpty(foundProvider.Class))
-                htmlAttributes.AddPropertyIfNotExist("class",
-                    (!string.IsNullOrEmpty(Options.Class) && !string.IsNullOrEmpty(foundProvider.Class))
-                    ? Options.Class + " " + foundProvider.Class
-                    : Options.Class + foundProvider.Class);
+            {
+                if (!string.IsNullOrEmpty(Options.Class))
+                    htmlAttributes.AddClass(Options.Class);
+                if (!string.IsNullOrEmpty(foundProvider.Class))
+                    htmlAttributes.AddClass(foundProvider.Class);
+            }
 
             htmlAttributes.AddPropertyIfNotExist("frameborder", "0");
             if (foundProvider.AllowFullScreen)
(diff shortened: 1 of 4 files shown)


src/Markdig/Extensions/MediaLinks/MediaLinkExtension.cs
```diff
@@ -154,10 +154,12 @@ private bool TryRenderIframeFromKnownProviders(Uri uri, bool isSchemaRelative, H
...
-                htmlAttributes.AddPropertyIfNotExist("class",
-                    (!string.IsNullOrEmpty(Options.Class) && !string.IsNullOrEmpty(foundProvider.Class))
-                    ? Options.Class + " " + foundProvider.Class
-                    : Options.Class + foundProvider.Class);
+            {
+                if (!string.IsNullOrEmpty(Options.Class))
+                    htmlAttributes.AddClass(Options.Class);
+                if (!string.IsNullOrEmpty(foundProvider.Class))
+                    htmlAttributes.AddClass(foundProvider.Class);
+            }
```
MihaZupan, line 162: Can you remove the surrounding `if` as it looks duplicated

RudeySH: Yep, done.

RudeySH:
Tests are failing because the class attribute is now appearing before the `width` and `height` attributes. I've already fixed this in [TestMediaLinks.cs](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/TestMediaLinks.cs) but I just noticed [MediaSpecs.md](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/MediaSpecs.md) and the unit test file that appears to get generated by that markdown file. When modifying these Specs .md files, how do I regenerate the *.generated.cs file along with it?

MihaZupan:
There is a SpecFileGen project in the solutuion. Running that will regenerate the test code

MihaZupan:
Can you also add a test case to emphasize the specific case this PR is fixing

RudeySH:
The thing is, Bootstrap adding `img-fluid` to embeds doesn't actually make sense. I'll add some tests but they'll have to be changed if https://github.com/lunet-io/markdig/issues/434 gets resolved.

RudeySH:
I'm currently running into https://github.com/nunit/nunit3-vs-adapter/issues/691 and I'm unable to run most tests in Visual Studio. I'll commit them but I have no idea if they'll pass, I guess we'll see.

MihaZupan:
Does it work if you run them from the command line with `dotnet test`?

RudeySH:
That works, thanks. It seems they pass!

Issue #451, opened 2020-07-21
Maths parsing problem


NaolShow:
Hi everyone, I have a problem with the parsing of the mathematics.
For exemple if I have this in markdown:

```
# Testing mathematics
$0$ or $2x$
```

The title will be parsed correctly, but I'll also have: $0$ or $2x$
(In the html source code: ```<p>$0$ or $2x$</p>```)

I'm using a pipeline:
```var Pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().UseBootstrap().UseEmojiAndSmiley().UseSoftlineBreakAsHardlineBreak().Build();```

Do you know what is the problem ?
Thanks a lot ! :)

xoofx:
Looks like a bug but can't recall what would be the cause of the problem. It seems to work fine with spaces around the number (e.g `$ 0 $`)
Don't have time to investigate issues these days for markdig, so if you feel to investigate this, you are welcome :)

davidackroyd99:
I will have a look and see if I can raise a pr

PR #452, opened 2020-07-21
Bug/parsing math

davidackroyd99:
This fixes [#451](https://github.com/lunet-io/markdig/issues/451)

Before my change the parser would look to see if the next character after the $ or $$ was a digit. It would then fail to match if it _was_ a digit, which means that `$x$` would be a valid math block but `$2x$` would not be. 

I simply removed this check as `$2x$` should be valid, but it still allows `$ 2x $` in order to be backwards compatible.

I also changed the WebApi project comment that specified that the second parameter should be **extensions**, where as actually it should just be **extension** as per the controller method definition. This confused me when I was trying to replicate the bug.

### DIFF
--- a/src/Markdig.WebApp/ApiController.cs
+++ b/src/Markdig.WebApp/ApiController.cs
@@ -13,7 +13,7 @@ namespace Markdig.WebApp
             return string.Empty;
         }
 
-        // GET api/to_html?text=xxx&extensions=advanced
+        // GET api/to_html?text=xxx&extension=advanced
         [Route("api/to_html")]
         [HttpGet()]
         public object Get([FromQuery] string text, [FromQuery] string extension)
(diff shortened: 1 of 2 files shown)


xoofx:
Thanks!

So I think I recall now that I replicated the behavior from pandoc described [here](https://pandoc.org/MANUAL.html#math) so that it should not start to parse `$20,000` as a mathematical expression... but as we can't have mixed spaces, so having `$20 followed by $50` would still work with your fix. It should be ok.

I will have a more thoroughly look at your PR later today

davidackroyd99:
I can add more tests to catch these edge cases and modify my code if needed?

xoofx:
It should be ok. I think that the rule of having previous separator/space before/after the $ sign should protect most of the cases. If a `$` is used as a currency prefix it will usually have a space before. That's likely the only rule used by pandoc, as I can see that in [the `This is $20 (20$)` case](https://pandoc.org/try/?text=This+is+%2420+(20%24)&from=markdown&to=html5&standalone=0) pandoc seems to interpret the whole as a math expression, which should be the same behavior now with this PR.

PR #453, opened 2020-07-21
Autolink tel: uri

davidackroyd99:
This fixes [437](https://github.com/lunet-io/markdig/issues/437)

Extends autolink functionality to be able to deal with tel:+1234567890.
Includes test.


xoofx:
Thanks!

Issue #454, opened 2020-07-30
[Pipe Tables Extension] Normalization is inserting extra columns


hamvocke:
I'm working with markdig's Pipe Tables Extension and found a case where the [specified behavior around normalization](https://github.com/lunet-io/markdig/blob/master/src/Markdig.Tests/Specs/PipeTableSpecs.md) is different from what [GitHub-flavored Markdown](https://github.github.com/gfm/#example-204) is specifying.

GFM says:

> The remainder of the table’s rows may vary in the number of cells. If there are a number of cells fewer than the number of cells in the header row, empty cells are inserted. If there are greater, the excess is ignored

markdig's pipe tables spec says:

> The tables are normalized to the maximum number of columns found in a table

---

This means, a table like this:

```
| abc | def |
| --- | --- |
| bar |
| bar | baz | boo |
```

is supposed to result in an HTML table that's got 2 columns according to GFM. The first body row gets an extra, empty `<td>` while the second body row doesn't include the `boo` cell.

Running this example through markdig with pipe tables enabled, however, generates a table that's got 3 columns, with the header and the first body row both gaining empty table cells.

I'm wondering if this deviation from GFM is intentional. Is there a reason why the two specifications treat normalization differently? And would it be desirable to make the normalization behavior configurable, e.g. by passing an option to the pipe tables plugin?

xoofx:
The GFM specs came officially much after Markdig pipe tables were developed. Before that was never really specified, probably more broken in terms of implicit specs than it is today. Never got the time to see how to get back to a closer behavior with GFM. A PR would be definitely welcome to do this work.

hamvocke:
That makes perfect sense, thanks for the explanation! Let me dig in and see what I can come up with to help here 👍

xoofx:
> Let me dig in and see what I can come up with to help here 👍

Good! So there is a major difference in the way we are parsing things in Markdig for pipe table ([here](https://talk.commonmark.org/t/parsing-strategy-for-tables/2027/49)), so the implementation today is a bit convoluted and not easy to work with. So don't be afraid about your first impression (e.g WTF is this code doing 😅 )

xoofx:
Meaning that, if we wanted to have 100% compatibility with GFM, the implementation would have to be quite different (but probably simpler). It could be done as a new pipe table parser mode, while keeping the existing implementation the default.

xoofx:
[...]

PR #455, opened 2020-07-30
Pipe Tables: Normalize using header column count

hamvocke:
This is a potential fix for #454.

In this PR I'm introducing `PipeTableOptions.UseHeaderForColumnCount`, an option that allows switching the PipeTables extension's normalization behavior. It's defaulting to the current behavior (find the row with the most columns and make sure all other rows have the same amount of cells). When setting this option to `true`, the normalization will follow the [example in GitHub's spec](https://github.github.com/gfm/#example-202) and remove/add cells in all rows until they match the number of cells in the header row.

I'm adding this as a draft PR for now since I need some guidance around the test suite:

Right now, I've changed the markdown example specs to conform to the new behavior and made the new behavior the default. Ideally, we'd have tests for both, the current and the new behavior - I just don't know how to do this. Does the markdown spec notation allow us to specify certain options to be used for the `MarkdownPipeline`?

### DIFF
--- a/src/Markdig.Tests/Specs/PipeTableGfmSpecs.generated.cs
+++ b/src/Markdig.Tests/Specs/PipeTableGfmSpecs.generated.cs
@@ -1,22 +1,24 @@
-// Generated: 2020-07-30 15:32:13
+// Generated: 2020-07-30 15:47:38
 
 // --------------------------------
-//            Pipe Tables
+//          GFM Pipe Tables
 // --------------------------------
 
 using System;
 using NUnit.Framework;
 
-namespace Markdig.Tests.Specs.PipeTables
+namespace Markdig.Tests.Specs.GFMPipeTables
 {
     [TestFixture]
-    public class TestExtensionsPipeTable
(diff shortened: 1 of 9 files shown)


xoofx:
So there is no way to specify this today easily, but you can maybe hack this at 

https://github.com/lunet-io/markdig/blob/bca544575bb47ba199bbd5d0f644152410fc0707/src/Markdig.Tests/TestParser.cs#L118-L121

Instead of making it part of Builder.Configure, you can early capture a special mode (`gfm-pipetable`) or add directly `gfm-pipetable` as a mode supported by Builder.Configure (could be ok in fact)

hamvocke:
Alright! I introduced a `gfm-pipetables` configuration for the pipeline builder and use that in a new `PipeTablesGfmSpec` (hence the huge diff all of a sudden). Don't know if this is overdoing it but this way we're ensuring both configurations pass all the pipe tables tests.

xoofx:
Perfect! Thanks for this enhancement and bringing it quickly!

PR #459, opened 2020-08-03
Write custom rel into links 

### LINKED ISSUE #456: rel='noopener' and rel='nofollow' as an additional option for links
Markdig already comes bundled with an extension – `.UseNoFollowLinks()` – that ensures all links have `rel='nofollow'` attribute. .

Meanwhile, `rel='noopener'` ensures that the target won't be able to access the window object of the page linking to it, which could be a security issue, and `rel='noreferrer'` ensures that the target page doesn't know where the visit came from. It can break analytics if used for internal links, but can be desired for user-generated content.

**Proposed solution A:**

Two more extensions, each adding its respective `rel` with `.UseNoOpenerLinks()` and `.UseNoReferrerLinks()`. It's a bit verbose and might have some impact on performance, but it's fully backwards-compatible.

**Proposed solution B:**

All three extensions merged into one, `.UseAdvancedLinks()`, possibly with something akin to the following signature:

```cs
UseAdvancedLinks(bool nofollow = false, bool noopener = false, bool noreferrer = false)
```

which could be used with named parameters: `.UseAdvancedLinks(noreferrer: true, nofollow: true)`. This cuts down on the verbosity and could be a bit more performant since all three attributes could be set in a single pass, but the issue might be backwards compatibility – should `.UseNoFollowLinks()` be dropped – or redundancy – should it exist besides `.UseNoFollowLinks()`.

I tried implementing it myself ([you can see my attempt here](https://github.com/Atulin/Markdig-Extensions)), but it turned out to be a task a bit too big for me. Perhaps some bigger brain will be able to implement it in a way that actually works.

iskcal:
This fixes #456 .

It allows writing custom rel into links. More tests are added.

I think there are two things that needs to be confirmed.

1. I consider that the name `RelXXX` is much suitable than `ReferralXXX` mentioned in #456 . For example, `UseRelLinks` has a straightfoward meaning to set rel property in links. If so, I will make another change.

2. Considering compatibility, I didn't remove the parts related to "nofollow" rel in links. Instead, I marked the parts as obsolete, such as `NoFollowLinksExtension`. I don't know whether it is an acceptable way.

### DIFF
--- a/src/Markdig.Tests/TestReferralLinks.cs
+++ b/src/Markdig.Tests/TestReferralLinks.cs
@@ -1,0 +1,71 @@
+using NUnit.Framework;
+
+namespace Markdig.Tests
+{
+    [TestFixture]
+    public class TestReferralLinks
+    {
+        [Test]
+        public void TestLinksWithNoFollowRel()
+        {
+            var markdown = "[world](http://example.com)";
+            var expected = "nofollow";
+
+#pragma warning disable 0618
+            var pipeline = new MarkdownPipelineBuilder()
+                .UseNoFollowLinks()
+#pragma warning restore 0618
+                .Build();
+            var html = Markdown.ToHtml(markdown, pipeline);
+
+            Assert.That(html, Contains.Substring($"rel=\"{expected}\""));
+        }
+
+        [Test]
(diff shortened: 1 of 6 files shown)


xoofx:
Thanks!

PR #466, opened 2020-08-16
Bring back net45 target

Kryptos-FR:
I am the author of [Markdig.Wpf](https://github.com/Kryptos-FR/markdig.wpf) and recently support for .NET framework was dropped in Markdig (https://github.com/lunet-io/markdig/pull/416). That is a huge breaking change for me since it means my library cannot be used anymore in traditional WPF appplication (i.e. not using netcore3.1 or later).

I propose to bring back at least net45 support since it doesn't need to many changes (only `Array.Empty<T>` need special care as of now).

I was also supporting net40 but it seems that it is already a done deal here. It was requested by the Visual Studio setup team since they are using my library (https://github.com/Kryptos-FR/markdig.wpf/issues/25). I have opened a discussion on my own repo to decide what to do for that net40 support (https://github.com/Kryptos-FR/markdig.wpf/issues/41).

*Note: `net452` is part of Windows 8.1 and will be supported until at least January 2023 (see https://support.microsoft.com/en-sg/lifecycle/search?alpha=Windows%208.1 and https://docs.microsoft.com/en-sg/lifecycle/faq/dotnet-framework#what-is-the-lifecycle-policy-for-different-versions-of-the-net-framework).*

### DIFF
--- a/src/Markdig.Tests/Markdig.Tests.csproj
+++ b/src/Markdig.Tests/Markdig.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-  	<TargetFrameworks>netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
+  	<TargetFrameworks>net452;netcoreapp2.1;netcoreapp3.1</TargetFrameworks>
     <OutputType>Library</OutputType>
     <IsPackable>false</IsPackable>
   </PropertyGroup>
--- a/src/Markdig.Tests/TestDescendantsOrder.cs
+++ b/src/Markdig.Tests/TestDescendantsOrder.cs
@@ -1,7 +1,7 @@
 using NUnit.Framework;
+using Markdig.Helpers;
 using Markdig.Syntax;
 using Markdig.Syntax.Inlines;
-using System;
 using System.Linq;
 using System.Collections.Generic;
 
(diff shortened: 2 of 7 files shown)


xoofx:
> I am the author of [Markdig.Wpf](https://github.com/Kryptos-FR/markdig.wpf) and recently support for .NET framework was dropped in Markdig (#416). That is a huge breaking change for me since it means my library cannot be used anymore in traditional WPF appplication (i.e. not using netcore3.1 or later).

Latest Markdig supports netstandard 2.0 which is compatible with .NET 4.7.1+ so saying that traditional WPF applications can't work anymore is not correct.

xoofx:
For net40, that will be a no go. A small OSS project like Markdig can't afford to have to maintain old FWK like net40.

Kryptos-FR:
I'm not talking about bringing back net40 (that is end of life anyway) but net45 is still supported until 2023.
As for .Net Standard this will also disappear when .Net 5 is released and not all companies have moved to .Net 4.7.

The main issue regarding 4.5 is that it is the default runtime installed on Windows 8.1. Default version on Windows 10 is also 4.6 so a huge portion of the market won't be able to use my library.

xoofx:
> but net45 is still supported until 2023

That's ok, the changes are minimal. But we will remove support for it at this date.

MihaZupan:
Quite unfortunate/ironic that Visual Studio is the one stuck on an ancient framework :/

Kryptos-FR:
> Quite unfortunate/ironic that Visual Studio is the one stuck on an ancient framework :/

According to [Visual Studio 2019 System Requirements](https://docs.microsoft.com/en-us/visualstudio/releases/2019/system-requirements#visual-studio-2019-system-requirements), it is the setup that is required to run on the stock version installed by default on the system. It does install .NET 4.7.2 as a part of the installation (same for [VS 2017](https://docs.microsoft.com/en-us/visualstudio/releases/2019/system-requirements#visual-studio-2019-system-requirements)).

Issue #469, opened 2020-08-21
Possible bug with string *bob*&_margaret_


xakpc:
Hey! 

I switched from CommonMark to Markdig to get more control over what markdown parse and what ignore and find out that one of my unit test failed.

Here is TestCase
`[TestCase("*bob*&_margaret_", "<em>bob</em>&amp;<em>margaret</em>")]`

CommonMark result: `<em>bob</em>&amp;<em>margaret</em>`
Markdig result: `<em>bob</em>&amp;_margaret_`

Is this a bug?

P.S. Check how it works in Github: *bob*&_margaret_

iskcal:
I have found the problem may be at the function `CheckUnicodeCategory` of class `CharHelper` from Line 174. The character & is not regarded as a valid punctuation in this function, because & may be used for HTML entities or print unicode. Markdig considers that the parts `&_margaret_` is a word and `_` has no effect in a word. If & is replaced to other punctuations, it would be parsed properly. For example, `*bob*+_margaret_` would be transformed into `<em>bob</em>+<em>margaret</em>`.

Interestingly, the code of this function seems to come from CommomMark.NET, but the behavior of CommonMark.NET is correct. If we don't want to change the function `CheckUnicodeCategory`, we can add a function to recoginize what & stands for, and it might be a good idea to see how CommonMark.NET implements it.

PR #470, opened 2020-08-25
Fix issue 469

iskcal:
Fix issue #469 . `&` can be a punctuation in function `CheckUnicodeCategory` of `CharHelper`. If `&` is a  beginner of a html entity or print unicode, it would be parsed by `HtmlEntityParser`.

### DIFF
--- a/src/Markdig/Helpers/CharHelper.cs
+++ b/src/Markdig/Helpers/CharHelper.cs
@@ -168,7 +168,7 @@ namespace Markdig.Helpers
             return IsZero(c) || IsWhitespace(c);
         }
 
-        // Check if a char is a pace or a punctuation
+        // Check if a char is a space or a punctuation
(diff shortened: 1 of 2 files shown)


src/Markdig/Helpers/CharHelper.cs
```diff
@@ -168,9 +168,7 @@ public static bool IsWhiteSpaceOrZero(this char c)
             return IsZero(c) || IsWhitespace(c);
         }
 
-        // Note that we are not considering the character & as a punctuation in HTML
-        // as it is used for HTML entities, print unicode, so we assume that when we have a `&` 
-        // it is more likely followed by a valid HTML Entity that represents a non punctuation
+        // Check if a char is a pace or a punctuation
```
KrisVandermotten, line 171: Typo

```suggestion
        // Check if a char is a space or a punctuation
```

iskcal: Thanks.

PR #471, opened 2020-08-28
Fix Setext headings in block quotes

### LINKED ISSUE #465: Setext headings in blockquotes
There seems to be a problem with [Setext headings](https://spec.commonmark.org/0.29/#setext-heading) used within blockquotes.

Take this markdown input:

```
> Hello
> ----
> this is some other text
```

Running this on [babelmark](https://babelmark.github.io/?text=%3E+Hello%0A%3E+----%0A%3E+this+is+some+other+text) shows that most renderers will generate a blockquote that contains a heading and a paragraph. Markdig, however, creates a blockquote that contains a paragraph, a horizontal rule and another paragraph.

The CommonMark spec is not absolutely explicit about this case but [generally states](https://spec.commonmark.org/0.29/#example-29) that `---` characters should be treated as a Setext heading if they can be interpreted that way.

Is this something that should be fixed in Markdig to be closer to what other Markdown renderers are doing?

hamvocke:
This PR attempts to fix #465 (spoiler alert: I've introduced another problem).

I've added a new test that makes sure that this markdown

```
> Hello
> ----
> this is some other text
```

is parsed as a block quote containing a `h2`. I made the test pass but introduced another regression around lazy continuations of block quotes. This is quite a head-scratcher for me so I'm getting this out with the failing test, maybe we can start discussing intermediate results and another pair of eyes helps fixing the remaining regression (or making sense of my overall approach)

### DIFF
--- a/src/Markdig.Tests/Specs/CommonMark.generated.cs
+++ b/src/Markdig.Tests/Specs/CommonMark.generated.cs
@@ -1,4 +1,4 @@
-// Generated: 2019-04-12 15:26:42
+// Generated: 2020-08-28 11:05:04
 
 // --------------------------------
 //        CommonMark v. 0.29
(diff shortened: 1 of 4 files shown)


xoofx:
Thanks for the PR! Sorry for not being able to review it quickly, I will try to come back to you later this week-end, pretty busy during this week.

This is indeed tricky, mostly because we currently don't know when we are in a continuation line or not, which makes this detection not easy to do.

xoofx:
Sorry, just back on this bug, I will try to have a look at it this week-end.
Discovered that it was no longer possible to debug the tests nicely because of a static constructor performing markdown parses, so I have fixed that.

xoofx:
This should be fixed by commit 8b64ce456f982a31da2c91bcea5ed052126a65dd

PR #481, opened 2020-10-03
Roundtrip implementation

generateui:
Implements a CST aka LST (Complete Syntax Tree aka Lossless Syntax Tree). 
Please keep the following in mind:
- some changes have broken Markdig parser
- tabs not yet implemented
- some hacky code here and there
- performance not taken into consideration/no perf tests run on this

As such, this is a very modest first attempt to learn if a CST can be implemented in markdig while keeping the current focus on performance. I'd be happy to hear feedback about
1. is having CST support desirable
2. what are pain points currently in included changes
3. any tips on moving forward

Any feedback is highly appreciated. 

For context, the usecase for this is creating bots that automatically enhance and fix markdown documentation. Think creating links automatically, fixing links automatically, inserting generated blocks of markdown code, et cetera. This requires a markdown renderer that does not change any input markdown as much as possible in the output markdown.


xoofx:
Indeed, Markdig was not built with that requirement but we have seen many cases where users would have loved to do roundtrip modifications to a Markdown document, so thanks for starting to work on this!

So far, the changes make senses. I will just add a couple of remarks and requirements:

- It's a huge amount of work because storing precisely spaces information (e.g need to store precisely which spaces characters were used) in every details will be a daunting detailed task. Expect several weeks of work, full-time. I also can't participate to the development due to limited sparetime availability for this project. I won't be able to finish the PR if it is stopped in the middle, but I should be able to review it once in a while you have made progress.
- Don't use the term Cst or Lst in the code. Similar to the fact that we are not using the term Ast in the code. 
- Focus on the CommonMark part before working on any extensions.
- The `NormalizeRenderer` would be the `MarkdownRenderer` (couldn't use this name because of the interface name and it would have been confusing) that performs no modifications on the Markdown. Any normalization could be done in a separate pass on the Ast.
- Because of performance, this feature must be optional. By default, when performing a Markdown to HTML, the code should not involve more allocations than it should. It will allocate a bit more for sure (many of the nodes will have to store for details), but we should try to minimize that. Ideally, this should not modify the performance drastically when turned off (e.g within 20%) while it can be e.g 2 times slower if it is turned on. It's important to keep that as a primary requirement because it will make your work more difficult but I wouldn't be able to merge it if we regress too much in terms of performance. This feature is only interesting for users that would like to do roundtrip modifications while the vast majority of users are just looking for the md2html feature.

I'm gonna add a few comments on the current PR. Ping me in the PR whenever you have made enough progress and you want another checkpoint.

Good luck with the work!

src/Markdig.Tests/TestCst.cs
```diff
@@ -0,0 +1,673 @@
...
+/// </summary>
+/// TODO:
+/// - \r\n, \r, \n
+/// - \t and spaces
+/// - html entities i.e. &gt;
+/// 
+namespace Markdig.Tests
+{
+    [TestFixture]
+    public class TestCst
```
xoofx, line 19: Don't use the term Cst. Prefer something like Roundtrip

src/Markdig.Tests/TestFencedCodeBlock.cs
```diff
@@ -0,0 +1,67 @@
...
+using NUnit.Framework;
+using System.IO;
+
+namespace Markdig.Tests
+{
+    [TestFixture]
+    public partial class TestFencedCodeBlock
+    {
+        [Test]
+        public void CstInfoParser_ParsesHappyFlow()
```
xoofx, line 12: Same no need for Cst term, you can remove it.

src/Markdig/Helpers/LineReader.cs
```diff
@@ -33,6 +33,11 @@ public LineReader(string text)
         /// </summary>
         public int SourcePosition { get; private set; }
 
+        public bool IsEndOfFile()
```
xoofx, line 36: Property?

src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -159,7 +166,27 @@ public BlockProcessor(MarkdownDocument document, BlockParserList parsers, Markdo
         /// <summary>
         /// Gets or sets a value indicating whether to continue processing the current line.
         /// </summary>
-        private bool ContinueProcessingLine { get; set; }
+        private bool _cpl;
```
xoofx, line 169: Don't think it is necessary to use _cpl here. Keep the full name, avoid acronyms in the code.

src/Markdig/Parsers/FencedBlockParserBase.cs
```diff
@@ -45,7 +45,7 @@ public abstract class FencedBlockParserBase<T> : FencedBlockParserBase where T :
         /// </summary>
         protected FencedBlockParserBase()
         {
-            InfoParser = DefaultInfoParser;
+            InfoParser = CstInfoParser;
```
xoofx, line 48: Keep the name DefaultInfoParser

src/Markdig/Parsers/NumberedListItemParser.cs
```diff
@@ -40,6 +40,7 @@ public override bool TryParse(BlockProcessor state, char pendingBulletType, out
                 {
                     startChar = endChar;
                 }
+                state.NextChar();
```
xoofx, line 43: What is this skipping? Add a comment if necessary

src/Markdig/Parsers/OrderedListItemParser.cs
```diff
@@ -37,6 +37,7 @@ protected bool TryParseDelimiter(BlockProcessor state, out char orderedDelimiter
             {
                 if (delimiter == orderedDelimiter)
                 {
+                    state.NextChar();
                     state.NextChar();
                     return true;
                 }
```
xoofx, line 43: What is this skipping? Add a comment if necessary

src/Markdig/Syntax/QuoteBlock.cs
```diff
@@ -12,6 +14,15 @@ namespace Markdig.Syntax
     /// <seealso cref="ContainerBlock" />
     public class QuoteBlock : ContainerBlock
     {
+        // ws? >? ws? block|inline
+        public class QuoteLine
```
xoofx, line 18: Avoid nested class. Move this above as a QuoteBlockLine or similar.

xoofx:
@MihaZupan don't hesitate to chime in if you have any comment!

generateui:
Thanks for the first glance and extensive feedback. In general I fully agree. 
1. Storing spaces is not just about parsing them right, but also deciding on what nodes they should be attached to. i.e. what to do on a line like `..>.>..-..text` (dots are spaces for clarity)? When writing this parser, most stuff is obvious yet some things can be really arbitrary or tricky. I have had a few times now where I unexpectedly had to change the node where to store whitespace on.
2. What's wrong with ast/cst? I agree that roundtrip conveys the meaning/intention better, but sometimes Cst/Lst seems to be a more accurate term.
3. Yes, extensions are second in prio
4. Yes, NormalizeRenderer is now abused to be the MarkdownRenderer. These classes will be copy/pasted and later some code deduplication will be done
5. Optionality is a goal, agreed. We must accept the fact that tree nodes will have properties with `null` values. I think such initial design will be fine, given these "cst/roundtrip properties" are documented well.

generateui:
Another thing I noticed during development is that changes in the parser sometimes produce broken trees. To have "lower-level tests" i.e. asserting that a given input produces a given tree structure, we'd need comparison implementation on nodes. This can be done on various ways, I wonder if you have put thought in what way is most desirable/appropriate for Markdig:
1. Implement `.Equals()` and `.HashCode()` with `.Equals()` overloads accepting a `ComparisonMethod { ContentAndStructure, Structure, WhitespaceContentAndStructure }` enum.
2. Implement IEqualityComparer multiple times, passing them into `Assert()` calls
3. Separate helper method classes
4. Convert tree to a specialized testing tree, and implement equality there

xoofx:
> Storing spaces is not just about parsing them right, but also deciding on what nodes they should be attached to. i.e. what to do on a line like ..>.>..-..text (dots are spaces for clarity)? When writing this parser, most stuff is obvious yet some things can be really arbitrary or tricky. I have had a few times now where I unexpectedly had to change the node where to store whitespace on.

Yes, that's not simple. If you take the way Roslyn C# parser is working, you have tokens e.g `>` and associated with them you can have trivias (e.g spaces, tabs...) before and after. When you are producing a model that want to be accurate to the source, you need to store also tokens (`>`) with their positions, and potentially the trivias associated to them (and their positions). 
And yes, that's lots of work. That's why I want you to realize that it's really several weeks of work!, just want to be sure that you know what you are "signing" for! 😅 

> What's wrong with ast/cst? I agree that roundtrip conveys the meaning/intention better, but sometimes Cst/Lst seems to be a more accurate term.

We don't need a term for that in the code base as I haven't used the term Ast in method names previously. Roundtrip was just for naming a test case. but I don't expect that we will have to rename many parsing method with "Roundtrip" or by adding a cst prefix. Let's keep it simple and not scrambled with terms like "cst" (also cst is more commonly understood as a shortcut for constant)

> To have "lower-level tests" i.e. asserting that a given input produces a given tree structure, we'd need comparison implementation on nodes. This can be done on various ways, I wonder if you have put thought in what way is most desirable/appropriate for Markdig:

If you want but yeah, it will require lots of tests to cover all these methods. Any code added should not lower the coverage that we have currently in our tests.

MihaZupan:
General question: do we need to keep the info on the type of whitespace in all these places?

Is the aim to have as close to lossless information in the syntax tree, or just to be able to reconstruct the structure of the document?

I'm not looking to start a tabs vs spaces debate, but commonmark has this to say about tabs

> Tabs in lines are not expanded to spaces. However, in contexts where whitespace helps to define block structure, tabs behave as if they were replaced by spaces with a tab stop of 4 characters.

> Thus, for example, a tab can be used instead of four spaces in an indented code block. (Note, however, that internal tabs are passed through as literal tabs, not expanded to spaces.)

Perf-wise, having to just store indentation counts where necessary would be better (that said I didn't measure the impact of these changes)

generateui:
As simple usecase, think of a bot parsing Markdown files in a folders, replacing links, and writing back markdown. These files are in a git repo, so when changes are comitted, you'll generally only want the actual changes. Accidental changes make the life of humans a lot harder when reviewing such changes or just digging into the logs.

From that perspective, you want to have character-for-character replication when you parse Markdown and then render back to Markdown again. I'm not sure this is actually possible, since in html and Markdown some quirky things happen. It should be possible to get quite close though.

In fact, one of my tests involves running the parse-render cycle on a bunch of Markdown files. A changelist then gives me all the problems still in the parser/renderer. As a nice side-effect, I suddenly see mistakes in markdown and authors not fully understanding Markdown. These last things can't be fixed with a roundtrip feature though ;)

MihaZupan:
That's a very good example.

For just links and some simple transformations, wouldn't it be a lot easier for you to use Regex replaces? I assume you had other, more complex transformations in mind?

src/Markdig/Syntax/Block.cs
```diff
@@ -48,6 +50,12 @@ protected Block(BlockParser parser)
         /// </summary>
         public bool RemoveAfterProcessInlines { get; set; }
 
+        public StringSlice BeforeWhitespace { get; set; }
+        public StringSlice AfterWhitespace { get; set; }
+
+        public List<StringSlice> LinesBefore { get; set; }
+        public List<StringSlice> LinesAfter { get; internal set; }
```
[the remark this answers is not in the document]
xoofx: I would expect that StringSlice are a better fit when working with an existing document, because you don't have to create multiple substrings of the original document to fill all the little whitespace blocks that we would have.

But when creating a new document from the API, each StringSlice would most likely be instantiated from plain separated strings.

generateui:
Thanks for the feedback so far. I have glanced over it, and placed handling it on the todolist (also added in the PR). Here and there, I have already resolved some of the detailed feedback.

generateui:
Progress update:
- All spec node types are now implemented. 
- Added test:
    - Block: ✔ 863 ❌ 11
    - Inline: ✔ 395 ❌ 6
    - Generated from examples: ✔ 649 ❌ 173
    - **Total**:  ✔ **1907** ❌ **190**
- See [todo/docs](https://github.com/generateui/markdig/blob/cst/src/Markdig.Tests/RoundtripSpecs/Roundtrip.md)

So what's next? Mopping up broken tests, cleaning the code and working feedback items.

src/Markdig.Tests/TestFencedCodeBlock.cs
generateui: done

src/Markdig/Helpers/LineReader.cs
generateui: removed unnecessary function

src/Markdig/Parsers/BlockProcessor.cs
generateui: fixed

src/Markdig/Parsers/NumberedListItemParser.cs
generateui: placed on todo

src/Markdig/Syntax/QuoteBlock.cs
generateui: done

src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -758,9 +807,16 @@ private bool TryOpenBlocks(BlockParser[] parsers)
 
                     if (!result.IsDiscard())
                     {
+                        // TODO: RTP: pass line with whitespace
                         paragraph.AppendLine(ref Line, Column, LineIndex, CurrentLineStartPosition);
                     }
-
+                    // TODO: RTP: delegate this to container parser classes
+                    var qb = paragraph.Parent as QuoteBlock;
+                    if (qb != null)
```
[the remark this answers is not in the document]
generateui: done

src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -159,7 +167,34 @@ public BlockProcessor(MarkdownDocument document, BlockParserList parsers, Markdo
...
+
+        public int WhitespaceStart { get; set; }
+
+        // NOTE: Line.Text is full source, not the line (!)
+        public StringSlice PopBeforeWhitespace(int end)
+        {
+            var stringSlice = new StringSlice(Line.Text, WhitespaceStart, end);
+            WhitespaceStart = end + 1;
+            return stringSlice;
+        }
```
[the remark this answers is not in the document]
generateui: removed unnecessary overload instead

src/Markdig/Renderers/Normalize/NormalizeRenderer.cs
```diff
@@ -50,15 +50,11 @@ public NormalizeRenderer(TextWriter writer, NormalizeOptions options = null) : b
 
         public bool CompactParagraph { get; set; }
 
-        public void FinishBlock(bool emptyLine)
+        public void FinishBlock(bool derp = false)
```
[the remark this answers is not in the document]
generateui: removed

src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -159,7 +167,34 @@ public BlockProcessor(MarkdownDocument document, BlockParserList parsers, Markdo
...
+
+
+                _cpl = value;
+            }
+        }
+
+        public int WhitespaceStart { get; set; }
+
+        // NOTE: Line.Text is full source, not the line (!)
+        public StringSlice PopBeforeWhitespace(int end)
```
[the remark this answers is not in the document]
generateui: done

src/Markdig/Syntax/Block.cs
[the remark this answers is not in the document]
generateui: removed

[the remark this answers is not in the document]
generateui: Indeed. StringSlice has a ctor accepting a String. If such apis will be used in the future, we may (carefully) consider an implicit operator from string, so api user can assign strings to these fields.

Is it safe to say we should use StringSlice types as general guideline for these fields primarily used for roundtrip parsing/generating?

generateui:
Progress update 2:
- TrackTrivia feature flag added
- most core classes adhere to TrackTrivia (BlockProcessor, InlineProcessor, etc)
- moved RoundtripRenderer to own folder, restored NormalizeRenderer
- link parsing uses 2 completely (duplicated for most part) code paths
- much cleaner code everywhere
- ❌ 14 Roundtrip tests failing
- ❌ 107 non-roundtrip tests failing
- ❌ **121 total** tests failing

Next steps: fix regressions. Most notable are list block parsing, fenced code blocks and StringLineGroup/StringSlice tests.

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -7,6 +7,44 @@
 
 namespace Markdig.Helpers
 {
+    /// <summary>
+    /// Wrap newline so we have type-safety and static accessibility
+    /// </summary>
+    public struct Newline
```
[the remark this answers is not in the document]
generateui: I was hoping to benefit from cases where we can pack this into two bits. May we benefit from 30 reduced bits as compared to an int enum?

src/Markdig.Tests/TestCst.cs
generateui: done

src/Markdig/Parsers/FencedBlockParserBase.cs
generateui: renamed to RoundtripInfoParser

src/Markdig/Parsers/OrderedListItemParser.cs
generateui: removed

src/Markdig/Parsers/FencedBlockParserBase.cs
```diff
@@ -59,6 +59,113 @@ protected FencedBlockParserBase()
...
+                        }
+                }
+            }
+
+            end:
+            fenced.WhitespaceAfterFencedChar = afterFence;
+            fenced.Info = HtmlHelper.Unescape(info);
+            fenced.WhitespaceAfterInfo = afterInfo;
+            fenced.Arguments = HtmlHelper.Unescape(arg);
+            fenced.WhitespaceAfterArguments = afterArg;
```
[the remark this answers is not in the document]
generateui: fixed by using StringSlice

src/Markdig/Parsers/MarkdownParser.cs
```diff
@@ -114,6 +114,15 @@ private void ProcessBlocks()
                 // If this is the end of file and the last line is empty
                 if (lineText.Text is null)
                 {
+                    if (!lineReader.IsEndOfFile())
+                    {
+                        var lastBlock = blockProcessor.LastBlock;
+                        if (lastBlock != null)
+                        {
+                            lastBlock.LinesAfter = blockProcessor.BeforeLines ?? new List<StringSlice>();
```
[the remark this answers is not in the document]
generateui: commented

src/Markdig/Syntax/Block.cs
```diff
@@ -48,6 +50,12 @@ protected Block(BlockParser parser)
         /// </summary>
         public bool RemoveAfterProcessInlines { get; set; }
 
+        public StringSlice BeforeWhitespace { get; set; }
+        public StringSlice AfterWhitespace { get; set; }
```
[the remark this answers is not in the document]
generateui: Unsure if I should document this here, in the code or in a separate documentation .md. I'll let this one open for now, until we resolve some discussion following from the PR.

src/Markdig/Syntax/Block.cs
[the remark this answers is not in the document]
generateui: I have now changed the current PR into using StringSlice everywhere where possible. Please let me know if this will resolve this item.

generateui:
I'm happy to announce that I have done all the todo items for this PR except benchmarking. This means that the current state of the code is -in principle- ready for actual merging. In practice, merging can not yet be performed due to:
- not yet benchmarked
- not yet reviewed
- some discussion points not yet resolved
- I may implement a testsuite which compares the tree where TrackTrivia is on/off, depending on how the discussion and testing will go
- some `TODO: RTP:` (intentionally) still present

But consider the current state as if the code will be merged. Thus, please perform a critical review on this PR. I will try to resolve these as quickly as possible. I regret merging the `main` branch into the `cst` branch, as the changes of `cst` and now combined with the recent changes. If this seriously hampers you I'll try to undo this merge.

Secondly, I'd like to open discussion on some points I encountered during implementation of this roundtrip parser. These are:
1. LinkHelper.cs currently has all functionality duplicated for roundtrip and non-roundtrip parsing. Should we keep this duplication? The core of the problem here is the amount of out parameters which make deduplication either hard or unwieldy. factoring out duplicated logic within these functions into private methods makes the (already hard to follow code) even harder to follow. I think I will definitely prefer the current state for the coming time.
2. The syntax tree will use StringSlice everywhere where possible.
3. I have added around 2000 tests. While this is a tremendous help, I can easily add around 1000 more test cases. With in-depth analysis, i may be able to go as far as 10K+. The question is: "will this detect more defects?". I'm personally quite satisfied with the current amount of tests covering most of the cases. The area where I'm less sure about is verification of the syntax tree model instance after parsing - there are some tests there, but the thoroughness is slim. When thinking around it, a standardized roundtrip/non-roundtrip test database may actually benefit other Commonmark implementations as well. I can whip up a model and move all these tests in there. If any interest is there from your side to do syntax tree instance validation, let me know, I'll provide a separate branch with guidance examples how I see this implemented.
4. Newlines. LineBreakInline now also parses `/r`, which I see as a drawback. It will most likely decrease performance as well. With the current API, this is not solvable, or at least not while inducing a lot of other non-elegant changes. This is because of how the ParserList class works and initialization of it when the MarkdownParser initializes. In other words, no dynamic `OpeningCharacters` can be returned in an `InlineParser` since the property is used at parsing initialization time.
5. Edgecases. Here and there, you will notice some special casing `if`s in the codebase. I tried to "wait them out" and hope for more cases following the same pattern, but some of them persisted. They are only a few, so I'd be inclined to keep them as-is but document them better.
6. TrackTrivia flag trickles down in around 6 classes. This is more of a comment for you guys to be aware of - I don't think we can more elegantly implement this as currently done. If you see ways on how to do this, let me know.
7. Roundtrip parsing documentation. Supporting Roundtrip is a massive feature which should be documented for extension implementors, syntax tree users (using the tree with TrackTrivia on and TrackTrivia off), tests and maintainers of this codebase. I have started documenting some small things in Roundtrip.md, but I have not yet put serious effort in this. I want to propose docs changes on the following places:
  1. **Roundtrip.md**: the nitty gritty details
  2. **Readme.md**: general support announcement and linking to Roundtrip.md
  3. **class level comments**: as currently done on classes and class members
  4. **tests**: as currently implemented in the three test suites
  5. **show-case tests**: new to-implement unit tests linked from Readme.md written with the developer in mind coming for the first time to the roundtrip feature. These tests will focus on showing the developer the general roundtrip functionality and as well demonstrating some nitty gritty details.

xoofx:
@generateui sorry to get back to you only now. I will try to have a look at your PR later this week-end

generateui:
Looking forward to your feedback!

xoofx:
Could you update your branch with latest master?

generateui:
done

src/Markdig/Helpers/StringSlice.cs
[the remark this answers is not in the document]
xoofx: > I was hoping to benefit from cases where we can pack this into two bits. May we benefit from 30 reduced bits as compared to an int enum?

You can force the underlying type of an enum to a byte, but the fact that this NewLine struct is then used within a class that will force padding anyway, so having a byte or an int doesn't much change the memory changes.

But the usage of an enum should simplifies the code.

src/Markdig/Syntax/IFencedBlock.cs
```diff
@@ -1,34 +1,99 @@
...
     {
+        /// <summary>
+        /// Gets or sets the fenced character used to open and close this fenced code block.
+        /// </summary>
+        char FencedChar { get; set; }
+
+        /// <summary>
+        /// Gets or sets the fenced character count used to open this fenced code block.
+        /// </summary>
+        public int OpeningFencedCharCount { get; set; }
```
xoofx, line 23: remove public modifier on interface

xoofx:
Thanks a lot for this PR, that's quite some work involved! 👍 

So I have run some benchmarks on the changes:

- The new version without trivias is 10% slower than previous, and takes 15% more memory
- The new version with trivias is 12% slower than without trivias and 12% more memory than without trivias

I haven't looked into the detail if we can improve anything, but I think it's acceptable in terms of performance impact considering the feature. @MihaZupan what do you think?

### Previous version

```
      Method |     Mean |     Error |    StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
------------ |---------:|----------:|----------:|---------:|---------:|---------:|----------:|
 TestMarkdig | 5.033 ms | 0.0296 ms | 0.0277 ms | 500.0000 | 250.0000 | 242.1875 |   2.88 MB |
```

### ParseTrivia off

```
      Method |     Mean |     Error |    StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
------------ |---------:|----------:|----------:|---------:|---------:|---------:|----------:|
 TestMarkdig | 5.494 ms | 0.0139 ms | 0.0108 ms | 507.8125 | 250.0000 | 242.1875 |   3.34 MB |
```

### ParseTrivia on

```
      Method |     Mean |     Error |    StdDev |    Gen 0 |    Gen 1 |    Gen 2 | Allocated |
------------ |---------:|----------:|----------:|---------:|---------:|---------:|----------:|
 TestMarkdig | 6.159 ms | 0.0356 ms | 0.0333 ms | 656.2500 | 257.8125 | 242.1875 |   3.76 MB |
```

xoofx:
> 1. LinkHelper.cs currently has all functionality duplicated for roundtrip and non-roundtrip parsing. Should we keep this duplication? The core of the problem here is the amount of out parameters which make deduplication either hard or unwieldy. factoring out duplicated logic within these functions into private methods makes the (already hard to follow code) even harder to follow. I think I will definitely prefer the current state for the coming time.

We should not complicate more, so I would tend to agree that we should let it as duplicated for now...


> 2. The syntax tree will use StringSlice everywhere where possible.

Good 👍 

> 3. I have added around 2000 tests. 

That's really good to have that many tests for the roundtrip. 

Btw, have you tried to run the roundtrip on the commonmark specs to see if it can actually output the same results?

> 4. Newlines. LineBreakInline now also parses /r, which I see as a drawback. It will most likely decrease performance as well. With the current API, this is not solvable, or at least not while inducing a lot of other non-elegant changes. This is because of how the ParserList class works and initialization of it when the MarkdownParser initializes. In other words, no dynamic OpeningCharacters can be returned in an InlineParser since the property is used at parsing initialization time.

I think it's ok. If you can change the NewLine to use an enum instead and that will fix the concern I have with it.

> 5. Edgecases. Here and there, you will notice some special casing ifs in the codebase. I tried to "wait them out" and hope for more cases following the same pattern, but some of them persisted. They are only a few, so I'd be inclined to keep them as-is but document them better.

I haven't checked.

> 6. TrackTrivia flag trickles down in around 6 classes. This is more of a comment for you guys to be aware of - I don't think we can more elegantly implement this as currently done. If you see ways on how to do this, let me know.

Yeah, that's probably ok, It's difficult to not flowing it if we want to avoid it.

> 7. Roundtrip parsing documentation. 

Agreed with your plan.

> 8. What to do with the \0 character?

In the specs we have:

> 2.3 Insecure characters
> For security reasons, the Unicode character U+0000 must be replaced with the REPLACEMENT CHARACTER (U+FFFD).

So, we should keep the replacement character in that case, even in a roundtrip scenario.

> 9. What to do with documents that have characters but no blocks i.e. \n \n?

Good question! Maybe if we have only trivias left we could add a fake ParagraphBlock with an single LiteralInline that has only trivias?

src/Markdig/Syntax/IBlock.cs
```diff
@@ -56,5 +57,9 @@ public interface IBlock : IMarkdownObject
         /// Occurs when the process of inlines ends for this instance.
         /// </summary>
         event ProcessInlineDelegate ProcessInlinesEnd;
+
+        public StringSlice WhitespaceBefore { get; set; }
+        public StringSlice WhitespaceAfter { get; set; }
```
MihaZupan, line 62: No public modifiers on the interface

MihaZupan (commented):
Thank you for an insane amount of work in this PR!

I skimmed through the code, haven't gone into too much details due to the sheer size.
We are heavily relying on the extensive test suite - speaking of which, thank you for adding a ton of tests.

I commented with some more questions / on a few things I believe **should change before** finalizing the PR.


> I haven't looked into the detail if we can improve anything, but I think it's acceptable in terms of performance impact considering the feature. @MihaZupan what do you think?

I saw a lot of things we can do to improve things further perf-wise, but nothing major that we should be blocking this PR on.

If the perf hit is ~10%, that is better than I expected, and surely acceptable considering the feature (Markdig's perf is insane as-is, I think it's just on some crazy people to push it futher as a hobby 😄).

I will have fun with perf in the weeks/months after this PR is merged.


> 8. What to do with the \0 character?

100% enforce replacement. We're heavily relying on \0 being special-cased everywhere. For example the CompactPrefixTree would break instantly if input contained null chars.


For a change this size, there are obiously a lot of nits that we could go into, but I agree with @xoofx that we shouldn't block and complicate this PR.

Thanks you

generateui:
Thanks for the nice words! Also, cool to see collab on the performance part. 

> Btw, have you tried to run the roundtrip on the commonmark specs to see if it can actually output the same results?

Yes, part of the PR is a generator generating Roundtrip tests using all the spec examples. They're run as part of the test suite. Around a third of the 2000 test cases come from this.

> I will have fun with perf in the weeks/months after this PR is merged.

That's awesome :)

Let's move forward on this PR. So, first let's cross off some todo-items:
1. ✔ leave `LinkHelper` as-is
2. ✔ `StringSlice`
3. ⚠ I came across a non-working SourceSpan on InlineLink, a crashing Roundtrip parser and a few other detailed unexpected states after parsing. If I find a structural issue, I won't be marking this PR to pull just yet, but do some prototypes on test generators
4. ✔ `/r`, `/r/n` in non-Roundtrip parser
5. ⚠ Edge cases: let's leave this one open until we do the merge. If you find them and conclude the're too ugly, I'll research better alternatives for them. If not, they'll be merged
6. ✔ trackTrivia flag trickling down
7. TODO docs
8. ✔ `\0`
9. TODO I suggest a separate block class for this one named `NoBlocksFoundBlock`

So, that leaves the current todo list:
1. docs
2. `NoBlocksFoundBlock`
3. handle review feedback
4. TODO: RTP
5. Assertion fail on a larger dataset
6. fix CI
7. optional: tree tests and IEqualityComparers
8. optional: dive into some edge cases

By the way, I believe the public-facing api is backward compatible. So that's something we should be good about.

iamcarbon:
@generateui -- Let me know I can help with any of this work to push this PR through. Was hoping to do another PR after this gets merged to enable nullable / but don't want to conflict.

generateui:
@iamcarbon:  I'm currently caught up in a lot of work, preventing me from finishing this PR in the coming time. It's still high on my prio list though, as I'm also playing around with Markdown. Having a roundtrip parser in Markdig really helps with these experiments :). So in short: help on this PR is very welcome, and I still plan to finalize it sometime soon.

src/Markdig/Renderers/Roundtrip/RoundtripRenderer.cs
```diff
@@ -0,0 +1,97 @@
...
+
+namespace Markdig.Renderers.Roundtrip
+{
+    /// <summary>
+    /// Default HTML renderer for a Markdown <see cref="MarkdownDocument"/> object.
+    /// </summary>
+    public class RoundtripRenderer : TextRendererBase<RoundtripRenderer>
+    {
+        /// <summary>
+        /// Initializes a new instance of the <see cref="NormalizeRenderer"/> class.
```
hawkerm, line 18: Default comments copied over, should update with more info for the new renderer.

generateui:
Update:

1. docs
2. ~~`NoBlocksFoundBlock`~~
3. ~~handle review feedback~~
4. ~~TODO: RTP~~
5. Assertion fail on a larger dataset (I cloned ~500 repos, and got some assertion fails when parsing all markdown)
6. fix CI
7. optional: tree tests and IEqualityComparers
8. optional: dive into some edge cases
9. 🆕: Rename whitespace to trivia where possible

generateui:
Update:

1. ~~docs~~
2. ~~`NoBlocksFoundBlock`~~
3. ~~handle review feedback~~
4. ~~TODO: RTP~~
5. Assertion fail on a larger dataset (I cloned ~500 repos, and got some assertion fails when parsing all markdown)
6. fix CI
7. optional: tree tests and IEqualityComparers
8. optional: dive into some edge cases
9. ~~🆕: Rename whitespace to trivia where possible~~
10. fix ContainerBlocksBlockQuotes_Example210

Getting **really** close!

src/Markdig.Tests/TestMediaLinks.cs
```diff
@@ -62,21 +62,21 @@ public TestHostProvider(string provider, string replace)
...
+        //{
+        //    string html = Markdown.ToHtml(markdown, GetPipeline(new MediaOptions
+        //    {
+        //        Hosts =
+        //        {
+        //            new TestHostProvider(provider, replace),
+        //        }
+        //    }));
+        //    Assert.AreEqual(html, expected);
+        //}
```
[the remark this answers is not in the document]
xoofx: @generateui don't forget to remove this test 😉

src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -712,18 +1069,19 @@ public static bool IsValidDomain(string link, int prefixLength)
                 segmentCount - lastUnderscoreSegment >= 2; // No underscores are present in the last two segments of the domain
         }
 
-        public static bool TryParseLinkReferenceDefinition<T>(T text, out string label, out string url,
-            out string title) where T : ICharIterator
-        {
-            return TryParseLinkReferenceDefinition(ref text, out label, out url, out title);
-        }
+        // apparently, below code is dead code (except for public api?)
```
xoofx, line 1072: Remove and commented code below if it is not used anymore

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -7,6 +7,50 @@
 
 namespace Markdig.Helpers
 {
+    /// <summary>
+    /// Wrap newline so we have type-safety and static accessibility
+    /// </summary>
+    public enum Newline
```
xoofx, line 13: Rename to `NewLine` and move it to its own separated file together with the NewLine extensions below.

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -220,10 +220,16 @@ internal static bool IsSpaceOrPunctuation(this char c)
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsNewLineLineFeed(this char c)
```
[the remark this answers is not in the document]
xoofx: Rename to `IsNewLineOrLineFeed`

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -220,10 +220,16 @@ internal static bool IsSpaceOrPunctuation(this char c)
...
+        {
+            return c == '\n' || c == '\r';
+        }
+
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static bool IsNewLine(this char c)
```
xoofx, line 230: This is confusing to say that IsNewLine returns true for `\r`, so I would prefer that we don't change the semantic here. Use `IsNewLineOrLineFeed` in place where it needs to support `\r` and `\n`

src/Markdig/Markdown.cs
```diff
@@ -153,13 +153,13 @@ public static MarkdownDocument Parse(string markdown)
         /// <param name="context">A parser context used for the parsing.</param>
         /// <returns>An AST Markdown document</returns>
         /// <exception cref="ArgumentNullException">if markdown variable is null</exception>
-        public static MarkdownDocument Parse(string markdown, MarkdownPipeline pipeline, MarkdownParserContext context = null)
+        public static MarkdownDocument Parse(string markdown, MarkdownPipeline pipeline, MarkdownParserContext context = null, bool trackTrivia = false)
```
xoofx, line 156: Actually, maybe we could add `TrackTrivia` as a property of `MarkdownPipeline`, and propagate the this bool to the processor when necessary?

src/Markdig/Parsers/FencedBlockParserBase.cs
```diff
@@ -45,7 +44,7 @@ public abstract class FencedBlockParserBase<T> : FencedBlockParserBase where T :
         /// </summary>
         protected FencedBlockParserBase()
         {
-            InfoParser = DefaultInfoParser;
```
xoofx, line 48: Shouldn't the default parser reroute internally to the previous default or round trip info parser depending on whether the `BlockProcessor`.`TrackTrivia` ?
Don't think that we want separated default + roundtripinfo if things are triggered by the presence of TrackTrivia

src/Markdig/Helpers/StringSlice.cs
generateui: As Wikipedia and Wiktionary suggest newline started as a compound word, but over time became a single word. I have moved Newline to a separate file. I strongly suggest to use Newline, but if you insist, I'll gladly make it NewLine.
https://en.wikipedia.org/wiki/Newline
https://en.wiktionary.org/wiki/newline

xoofx: The .NET framework is using names like `Environment.NewLine`, so I would like that this to be consistent.

generateui: That's fair, I'll change it to NewLine on the class and member names.

src/Markdig/Parsers/FencedBlockParserBase.cs
generateui: fixed, thanks for spotting. This should improve perf slightly.

src/Markdig/Markdown.cs
generateui: done

generateui:
Update:

1. ~~docs~~
2. ~~`NoBlocksFoundBlock`~~
3. ~~handle review feedback~~
4. ~~TODO: RTP~~
5. ~~Assertion fail on a larger dataset (I cloned ~500 repos, and got some assertion fails when parsing all markdown)~~
6. ~~fix CI~~
7. optional: tree tests and IEqualityComparers
8. optional: dive into some edge cases
9. ~~🆕: Rename whitespace to trivia where possible~~
10. ~~fix ContainerBlocksBlockQuotes_Example210~~

🎉🥳 DONE! As far as I can see, all items are done and this PR is ready to merge.

MihaZupan:
I'll do a once-over tomorrow.

src/Markdig/Parsers/Inlines/EscapeInlineParser.cs
```diff
@@ -41,20 +41,47 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
...
+                        IsHard = true,
+                        IsBackslash = true,
+                        Span = { Start = processor.GetSourcePosition(startPosition, out line, out column) },
+                        Line = line,
+                        Column = column
+                    };
+                    processor.Inline.Span.End = processor.Inline.Span.Start + 1;
+                    slice.NextChar();
+                    return true;
+                }
```
MihaZupan, line 82: We're not really saving anything by duplicating the code here, we can skip the TrackTrivia
 check here

src/Markdig/Parsers/QuoteBlockParser.cs
```diff
@@ -56,21 +90,62 @@ public override BlockState TryContinue(BlockProcessor processor, Block block)
...
+                }
             }
-
-            c = processor.NextChar(); // Skip opening char
-            if (c.IsSpace())
+            c = processor.NextChar(); // Skip quote marker char
+            if (c == ' ')
             {
-                processor.NextChar(); // Skip following space
+                //processor.NextChar();
```
MihaZupan, line 118: Should this be removed?

src/Markdig/Renderers/Roundtrip/CodeBlockRenderer.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace Markdig.Renderers.Roundtrip
+{
+    /// <summary>
+    /// An Normalize renderer for a <see cref="CodeBlock"/> and <see cref="FencedCodeBlock"/>.
+    /// </summary>
+    /// <seealso cref="NormalizeObjectRenderer{CodeBlock}" />
+    public class CodeBlockRenderer : RoundtripObjectRenderer<CodeBlock>
+    {
+        public bool OutputAttributesOnPre { get; set; }
```
MihaZupan, line 16: This looks unused

src/Markdig/Renderers/Roundtrip/Inlines/LineBreakInlineRenderer.cs
```diff
@@ -0,0 +1,30 @@
...
+    /// <summary>
+    /// A Normalize renderer for a <see cref="LineBreakInline"/>.
+    /// </summary>
+    /// <seealso cref="NormalizeObjectRenderer{LineBreakInline}" />
+    public class LineBreakInlineRenderer : RoundtripObjectRenderer<LineBreakInline>
+    {
+        /// <summary>
+        /// Gets or sets a value indicating whether to render this softline break as a Normalize hardline break tag (&lt;br /&gt;)
+        /// </summary>
+        public bool RenderAsHardlineBreak { get; set; }
```
MihaZupan, line 18: Looks unused

src/Markdig/Renderers/Roundtrip/ThematicBreakRenderer.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        protected override void Write(RoundtripRenderer renderer, ThematicBreakBlock obj)
+        {
+            renderer.RenderLinesBefore(obj);
+
+            // for now, render always a newline
+            // TODO: only render a newline when not last line
+            //renderer.Write(obj.BeforeWhitespace);
+            renderer.Write(obj.Content);
+            //renderer.Write(obj.AfterWhitespace);
```
MihaZupan, line 23: Should these be removed?

src/Markdig/Syntax/IBlock.cs
```diff
@@ -56,5 +57,9 @@ public interface IBlock : IMarkdownObject
         /// Occurs when the process of inlines ends for this instance.
         /// </summary>
         event ProcessInlineDelegate ProcessInlinesEnd;
+
+        public StringSlice TriviaBefore { get; set; }
+        public StringSlice TriviaAfter { get; set; }
```
MihaZupan, line 62: no `public` modifier on interfaces

MihaZupan (commented):
Left a few comments.

There's a few outstanding comments that should also be resolved
https://github.com/xoofx/markdig/pull/481#discussion_r546815957
https://github.com/xoofx/markdig/pull/481#discussion_r586101206 (missing `Or`)
https://github.com/xoofx/markdig/pull/481#discussion_r546741666
https://github.com/xoofx/markdig/pull/481#discussion_r579970564

`Newline.cs` should be renamed to `NewLine.cs`, `NoBlocksFoundBlock.cs` to `EmptyBlock.cs`

generateui:
Thanks for taking a look again, MihaZupan! There were quite some small things I could fix for the better.

Let's merge this one :)

xoofx:
It's merged! Thanks for all the massive work @generateui that you put on this PR, really appreciated.

generateui:
> It's merged! Thanks for all the massive work @generateui that you put on this PR, really appreciated. 👍

w00t! It was a fun and lengthy challenge. Thanks for your collab, @xoofx and @MihaZupan!

Issue #487, opened 2020-11-01
No easy access to label for links that use reference links


ErikSchierboom:
When working with inline links, the behavior of the `LabelSpan` property seems to be inconsistent between inline links that use reference links and those that do not.

This is best demonstrated with an example.

Suppose we're parsing the following Markdown

```markdown
[Link using reference][valid-file]
[Link without reference](/valid-file-link.md)

[valid-file]: ./valid-file-link.md
```

using the following code:

```csharp
var pipeline = MarkdownPipelineBuilder().UsePreciseSourceLocation().Build();
var markdownDocument = Markdown.Parse(markdown, pipeline);
```

Then if I use the Span, LabelSpan and UrlSpan properties to output the corresponding Markdown in the document, I get the following:

First link (using the reference link):
- Span: `"[Link using reference][valid-file]"`
- LabelSpan: `"valid-file"`
- UrlSpan: `"./valid-file-link.md"`

Second link (not using the reference link):
- Span: `"[Link without reference](/valid-file-link.md)"`
- LabelSpan: `"Link without reference"`
- UrlSpan: `"/valid-file-link.md"`

I would have expected the `LabelSpan` property of the first link to return `Link using reference`, not `valid-file`.

Am I doing something wrong or is this expected behavior?

MihaZupan:
The behavior looks wrong here. We're also not setting the `Label` for `[foo](bar)` case.

ErikSchierboom:
@MihaZupan I've opened PR #488 to fix this issue. Is there anything I should be doing to get the PR reviewed?

PR #488, opened 2020-11-01
Fix label of inline links not set correctly

ErikSchierboom:
Closes https://github.com/lunet-io/markdig/issues/487

I'm new to the code base, so I have no idea if this is the right approach. It does fix the issue. I didn't know where to add a test for this. Let me know if you'd like me to add such a test.

### DIFF
--- a/src/Markdig.Tests/TestLinkParser.cs
+++ b/src/Markdig.Tests/TestLinkParser.cs
@@ -20,18 +20,18 @@ namespace Markdig.Tests
         [Test]
         public void TestReferenceLink()
         {
-            var doc = Markdown.Parse("There is a [link][foo]\n\n[foo]: /toto");
+            var doc = Markdown.Parse("There is a [ref][foo]\n\n[foo]: /toto");
             var link = doc.Descendants<LinkInline>().First();
-            Assert.AreEqual("link", link.Label);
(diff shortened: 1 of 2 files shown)


xoofx:
Could you add a small `TestLinkParser.cs` with a test extracted from the issue #487?

ErikSchierboom:
@xoofx Sure! My pleasure.

ErikSchierboom:
@xoofx I've found an issue with my implementation, so it could take a little longer!

ErikSchierboom:
@xoofx Apparently I've misread the spec 🤦 The Label is used for the reference definition linking.

After some trial and error, I've found that I can get the link text using the following method (having enabled precise source location):

```csharp
string GetIdentifier(LinkInline link, string markdown)
{
    if (link.IsShortcut)
    {
         return link.Label;
    }

    return markdown[link.FirstChild.Span.Start..(link.LastChild.Span.End + 1)];
}
```

Do you think there is any value in adding a `TextSpan` property to the `LinkInline` class to allow the above code to be simplified:

```csharp
string GetIdentifier(LinkInline link, string markdown)
{
    return markdown[link.TextSpan.Start..(link.TextSpan.End + 1)];
}
```

Let me know what you think. If you don't think this is valuable, feel free to close this issue.

PR #492, opened 2020-11-10
Added a reference to WebStoating.Markdig.Prism

ilich:
Hello,

I have added a reference to Markdig extensions that renderers various programming languages source code using Prism.js.


xoofx:
Thanks!

Issue #180, opened 2017-12-05
Improve performance for large tables


xoofx:
As seen in the issue #179 a large pipe table doesn't scale with the current code. Need to figure out how to optimize it.

MihaZupan:
I think we can count this as done with #922

PR #500, opened 2021-01-07
Add depth limits to avoid pathological-case parsing times/StackOverflows

### LINKED ISSUE #497: StackOverflowException is throw when converting special markdown to HTML
We use Markdig in our code to convert Markdown to HTML. It works perfect in most cases w.r.t. to performance. Recently we received reports that the library hangs and throws StackOverflowException when processing some special markdowns.

Here are the sample code:
`string input = <special markdown input>;
var pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().Build();

string result = Markdown.ToHtml(input, pipeline);`

@xoofx, I wouldn't paste the markdown payload here due to security reason, please let me know the best option to share it out to continue investigation.

MihaZupan:
Fixes #497 

Due to the recursive approach to rendering, a very deep AST would trigger a StackOverflow during rendering.
This PR adds a general depth limit used in the rendering phase to detect deep nesting (currently 128 which can't affect any non-malicious inputs).
This shouldn't be problematic at all, offering a cheap fail-fast and StackOverflow protection.

Some inputs hit trivial O(n^2) cases due to depth - `new string('[', 100_000)` or `new string('>', 100_000)`.
It also adds a large depth limit used to fail-fast such inputs that would otherwise take ***minutes*** to process.
The current limit of `10 * 1024` is rather arbitrary - on my machine it means that such inputs will fail after ~200 ms.
For comparison, 8192 results in ~125 ms and 16384 in ~500 ms.
**Note: the problem with this limit is that it limits the maximum size of a pipe table due to the way the parser is implemented (#180) - ~10k table cells.**

Should the second part (large limit also affecting tables) be configurable?
We may be able to avoid the table limitation in the future if we revisit the table parsing logic.

CC: @leonyang12

Users can semi-easily avoid the rendering StackOverflow by analyzing the AST after the parsing step (https://gist.github.com/MihaZupan/2e25659bc9af623589df1d1cb1ce712c).

Without a change directly in Markdig, I don't think avoiding the very long `Parse` times is practical.

This PR does **not guarantee** that ALL possible malicious-like inputs will fail-fast.

### DIFF
--- a/src/Markdig/Helpers/ThrowHelper.cs
+++ b/src/Markdig/Helpers/ThrowHelper.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
+using System.Runtime.CompilerServices;
 
 namespace Markdig.Helpers
 {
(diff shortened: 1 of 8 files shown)


src/Markdig/Helpers/ThrowHelper.cs
```diff
@@ -34,6 +35,28 @@ internal static class ThrowHelper
 
         public static void InvalidOperationException(string message) => throw new InvalidOperationException(message);
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void CheckDepthLimit(int depth, bool useLargeLimit = false)
+        {
+            // Very conservative limit used to limit nesting in the final AST
+            // Used to avoid a StackOverflow in the recursive rendering process
+            const int DepthLimit = 128;
```
leonyang12, line 43: Can we make this configurable? We might want to set a pretty small depth in our case

leonyang12 (approved):
Looks good. Thanks for the fix.

xoofx:
Thanks!

PR #502, opened 2021-01-12
Fix CodeInline to plain text incorrect escape

### LINKED ISSUE #495: Bug: HtmlRenderer incorrectly renders `>`
When set `EnableHtmlEscape = false`, the ">" is still be escaped and output is empty string: "". Looks it is an issue of `CodeInlineRenderer` class.

yufeih:
Fixes #495 

This PR fixes `CodeInlineRenderer` the same way as [LiteralInlineRenderer](https://github.com/xoofx/markdig/blob/master/src/Markdig/Renderers/Html/Inlines/LiteralInlineRenderer.cs) by checking `EnableHtmlEscape` and call different methods accordingly.


xoofx:
Thanks!

PR #511, opened 2021-02-09
Expose `TextRendererBase.Reset` method to inheritors

### LINKED ISSUE #508: Ability to cache and reuse HtmlRenderer instances
There are some scenarios requiring ability to implement a custom cache of `HtmlRenderer` objects.
One scenario is when a `MarkdownDocument` needs to be parsed and processed first before rendering it into a string:

```cs
// 1. Parse and process a Markdown document:
var document = Markdown.Parse(text, _pipeline);
/* process the document here */

// 2. Render the document to a string:
//      it'd be nice to use a cached renderer,
//      instead of creating a new one each time.
var writer = new StringWriter();
var renderer = new HtmlRenderer(writer);
_pipeline.Setup(renderer);

var renderedStr = renderer.Render(document);
// ...
```

Another scenario includes ability to create and cache `HtmlRenderer` objects having custom settings. Like the ones stripping Html from inlines implemented [here](https://github.com/xoofx/markdig/blob/3030b72f781f902cb947a173853de95de8924c6c/src/Markdig/Extensions/AutoIdentifiers/AutoIdentifierExtension.cs#L197-L202).

The problem with caching is that a `HtmlRenderer` object needs to be reset after usage, but the [`TextRendererBase.Reset`](https://github.com/xoofx/markdig/blob/3030b72f781f902cb947a173853de95de8924c6c/src/Markdig/Renderers/TextRendererBase.cs#L93-L107) method is declared as `internal`, what makes reuse impossible. Do you think accessibility of the method can be revised to allow `protected`, or even `public` access?

AndreyChechel:
Fixes #508 by exposing the [`TextRendererBase.Reset`](https://github.com/xoofx/markdig/blob/3030b72f781f902cb947a173853de95de8924c6c/src/Markdig/Renderers/TextRendererBase.cs#L93-L107) method to the class inheritors.
That change makes it possible to implement custom caches of `HtmlRenderer` instances.

### DIFF
--- a/src/Markdig/Renderers/TextRendererBase.cs
+++ b/src/Markdig/Renderers/TextRendererBase.cs
@@ -90,7 +90,7 @@ namespace Markdig.Renderers
             indents = new List<string>();
         }
 
-        internal void Reset()
+        protected internal void Reset()
         {
             if (Writer is StringWriter stringWriter)
             {


MihaZupan (approved):

src/Markdig/Renderers/TextRendererBase.cs
```diff
@@ -90,7 +90,7 @@ protected TextRendererBase(TextWriter writer) : base(writer)
             indents = new List<string>();
         }
 
-        internal void Reset()
```
xoofx, line 93: Wondering if we shouldn't rename this method to `ResetInternal` that would call a `protected virtual void Reset()` method instead?
@MihaZupan ?

MihaZupan: I don't think we need the internal method in that case.

`protected internal virtual void Reset()`?

I've never seen that signature, but apparently it's legal.

MihaZupan: Though I'm not sure what we'd be getting from making it virtual - we would never call it on renderers we don't own (since the writer might not support it)

AndreyChechel: From what I can see, the `Reset` method is always invoked by the code having control over `HtmlRenderer` instantiation.
So there is no point in making the method virtual - basically what @MihaZupan said in the previous comment.

Custom `HtmlRenderer` inheritors will be able to hide the method with a virtual one, if they need virtual behavior for their derived types.

PR #519, opened 2021-03-07
ADD: Markdown.ToHtml from MarkdownDocument

grishat:
Hello. In my case, I had to add additional classes in html elements. And the best way, in my opinion, was to add attributes on parsed document, and then render this one. And I added this missed method.
Code in my local project:
```
if (inline is LinkInline link)
{
   link.GetAttributes()
      .AddClass("mud-typography mud-link mud-primary-text mud-link-underline-hover mud-typography-body1");
}
```

### DIFF
--- a/src/Markdig/Markdown.cs
+++ b/src/Markdig/Markdown.cs
@@ -93,7 +93,7 @@ namespace Markdig
         /// <param name="pipeline">The pipeline used for the conversion.</param>
         /// <returns>The result of the conversion</returns>
         /// <exception cref="ArgumentNullException">if markdown document variable is null</exception>
-        public static string ToHtml(MarkdownDocument document, MarkdownPipeline pipeline = null)
+        public static string ToHtml(this MarkdownDocument document, MarkdownPipeline pipeline = null)
         {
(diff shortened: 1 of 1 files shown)


src/Markdig/Markdown.cs
```diff
@@ -86,6 +86,28 @@ public static string ToHtml(string markdown, MarkdownPipeline pipeline = null, M
...
         }
 
+        /// <summary>
+        /// Converts a Markdown document to HTML.
+        /// </summary>
+        /// <param name="document">A Markdown document.</param>
+        /// <param name="pipeline">The pipeline used for the conversion.</param>
+        /// <returns>The result of the conversion</returns>
+        /// <exception cref="ArgumentNullException">if markdown document variable is null</exception>
+        public static string ToHtml(MarkdownDocument document, MarkdownPipeline pipeline = null)
```
xoofx, line 96:
```suggestion
        public static string ToHtml(this MarkdownDocument document, MarkdownPipeline pipeline = null)
```

Maybe make it extension methods as well by adding this?

grishat: Yes, it looks better.

PR #520, opened 2021-03-07
Some minor cleanup

MihaZupan:


MihaZupan:
cbd00a4 while unfortunate seems necessary. Alternatively, the example would be a lot uglier (passing explicit new lines).

src/Markdig/Renderers/Roundtrip/CodeBlockRenderer.cs
```diff
@@ -2,6 +2,7 @@
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
 
+using Markdig.Renderers.Normalize;
```
xoofx, line 5: Do you know why this namespace is added?

src/Markdig/MarkdownPipeline.cs
```diff
@@ -52,7 +52,7 @@ internal MarkdownPipeline(OrderedList<IMarkdownExtension> extensions, BlockParse
         /// True to parse trivia such as whitespace, extra heading characters and unescaped
         /// string values.
         /// </summary>
-        public bool TrackTrivia { get; set; }
+        internal bool TrackTrivia { get; set; }
```
xoofx, line 55: Leave it public getter - otherwise once building a pipeline, we can't know if it's asking trivia or not.

src/Markdig/MarkdownPipelineBuilder.cs
```diff
@@ -80,7 +80,7 @@ public MarkdownPipelineBuilder()
         /// True to parse trivia such as whitespace, extra heading characters and unescaped
         /// string values.
         /// </summary>
-        public bool TrackTrivia { get; set; }
+        internal bool TrackTrivia { get; set; }
```
xoofx, line 83: How to setup the trivia on builder if we can't access it from the public API?

Probably needs an extension method on the builder e.g `WithTrivia()`

src/Markdig/Syntax/Inlines/LinkInline.cs
```diff
@@ -3,6 +3,7 @@
 // See the license.txt file in the project root for more information.
 
 using Markdig.Helpers;
+using Markdig.Parsers;
```
xoofx, line 6: Why this is added? - same above -

xoofx: Oh, maybe for XML documentation?

src/Markdig/Renderers/Roundtrip/CodeBlockRenderer.cs
xoofx: Oh, maybe it is for XML documentation. I haven't looked at the code, but I'm curious what dependency we have with normalize...

src/Markdig/MarkdownPipelineBuilder.cs
MihaZupan: There is a `EnableTrackTrivia` extension exposed, but it does more than just set this bool. The actual behavior can't be toggled on/off with just this bool, so having it introduces ambiguity where only the extension method should be used.

src/Markdig/Renderers/Roundtrip/CodeBlockRenderer.cs
MihaZupan: It was for the XML docs. The issue was that the comment was referencing the NormalizeRenderer (as that was used as a base for the roundtrip). I removed these usings and corrected the docs to point to the RoundtripRenderer.

src/Markdig/Syntax/Inlines/LinkInline.cs
MihaZupan: Also for XML documentation. I double-checked, and it was only needed for a reference to `MarkdownParser.TrackTrivia`. Since that class is never actually exposed to the user as an instance, I changed these to point to `MarkdownPipeline.TrackTrivia` instead.

src/Markdig/MarkdownPipelineBuilder.cs
xoofx: Fair enough, but maybe make it public getter still - as you might want to develop an extension that can react to a previously setup track trivia.

MihaZupan: Changed to

```c#
public bool TrackTrivia { get; internal set; }
```

PR #499, opened 2020-12-17
Random perf improvements

MihaZupan:
Codegen tweaks and allocation improvements.

Removing the overheads in the simple `Markdown.ToHtml(...)` case with the default pipeline.
For an empty string allocations go from 17.14 KB to 928 B and take 1/20th the time because of the default pipeline caching (2af8b1f) - means we're now benefiting from the Renderer caching from https://github.com/lunet-io/markdig/commit/afe4308e918b8afa5d5359d2fab270b4621ef3ab in the basic use cases.

For c776d18 I was quite surprised how bad the compiler is at avoiding these allocations :/

PR #521, opened 2021-03-07
Random perf improvements

MihaZupan:
Brings over some of the changes from #499:
- Caching the default pipeline
- Caching for custom writers
- Reduced MarkdownObject size
- Removed most heap allocations from ProcessInlines loop

### DIFF
--- a/src/Markdig/Renderers/TextRendererBase.cs
+++ b/src/Markdig/Renderers/TextRendererBase.cs
@@ -28,9 +28,7 @@ namespace Markdig.Renderers
         protected TextRendererBase(TextWriter writer)
         {
             if (writer == null) ThrowHelper.ArgumentNullException_writer();
-            this.Writer = writer;
-            // By default we output a newline with '\n' only even on Windows platforms
-            Writer.NewLine = "\n";
+            Writer = writer;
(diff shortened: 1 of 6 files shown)


src/Markdig/Renderers/TextRendererBase.cs
```diff
@@ -128,6 +128,11 @@ protected internal void Reset()
                 ThrowHelper.InvalidOperationException("Cannot reset this TextWriter instance");
             }
 
+            ResetInternal();
+        }
+
+        internal void ResetInternal()
```
MihaZupan, line 134: There is now a split between an exposed `Reset` and internal `ResetInternal`.

```c#
protected internal void Reset()
{
    // Reset Writer

    ResetInternal();
}

internal void ResetInternal()
{
    // Reset internal state
}
```

We only use `ResetInternal` in the library as we can't reset an arbitrary writer.

Would it make sense to change the exposed `Reset` to not reset the writer and leave that to the caller instead?

xoofx: Don't have enough insights on the implication, so I trust your choice 😉

PR #522, opened 2021-03-08
Enable nullability

iamcarbon:
This PR aims to enable nullability for the markdig project. To avoid conflicts with existing PR's, it's proposed that nullability be enabled incrementally on a per-file basis before enabling for the project and removing the directives. This effort is likely to span multiple PRs.


xoofx:
Thanks!

PR #524, opened 2021-03-10
Enable nullable annotations, round 2 of x

iamcarbon:
Round 2

### DIFF
--- a/src/Markdig/Extensions/Tables/PipeTableParser.cs
+++ b/src/Markdig/Extensions/Tables/PipeTableParser.cs
@@ -2,6 +2,8 @@
 // This file is licensed under the BSD-Clause 2 license.
 // See the license.txt file in the project root for more information.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
@@ -28,7 +30,7 @@ namespace Markdig.Extensions.Tables
         /// </summary>
         /// <param name="lineBreakParser">The linebreak parser to use</param>
         /// <param name="options">The options.</param>
-        public PipeTableParser(LineBreakInlineParser lineBreakParser, PipeTableOptions options = null)
+        public PipeTableParser(LineBreakInlineParser lineBreakParser, PipeTableOptions? options = null)
         {
             this.lineBreakParser = lineBreakParser ?? throw new ArgumentNullException(nameof(lineBreakParser));
             OpeningCharacters = new[] { '|', '\n', '\r' };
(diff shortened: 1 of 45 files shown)


xoofx:
I had to merge another PR in the meantime and there are a few merge conflicts, could you have a look?

src/Markdig/Extensions/Tables/PipeTableParser.cs
```diff
@@ -59,7 +61,7 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             var position = processor.GetSourcePosition(slice.Start, out int globalLineIndex, out int column);
             var localLineIndex = globalLineIndex - processor.LineIndex;
 
-            if (tableState == null)
+            if (tableState is null)
```
MihaZupan, line 64: I don't mind these changes (personally prefer is/is not), but it doesn't make a difference for nullable analyzers right?

src/Markdig/Parsers/ParserList.cs
```diff
@@ -60,11 +61,11 @@ protected ParserList(IEnumerable<T> parsersArg) : base(parsersArg)
             var tempCharMap = new Dictionary<char, T[]>();
             foreach (var parser in this)
             {
-                if (parser.OpeningCharacters != null && parser.OpeningCharacters.Length != 0)
+                if (parser.OpeningCharacters is { Length: > 0 })
```
MihaZupan, line 64: We should stay consistent with the type of checks. The same pattern is used above in the same method but wasn't changed.

MihaZupan:
Thanks, I think this is great for Markdig. Even if we might end up with a ton of `!`s in code rn, it's a great start.

src/Markdig/Parsers/ParserList.cs
iamcarbon: Agreed. I had planned to follow up to apply these changes everywhere, but didn't want this PR to become to big. I replaced the remaining instances of this pattern to address this feedback now.

src/Markdig/Extensions/Tables/PipeTableParser.cs
iamcarbon: The analyzers pick up both patterns, but is is slightly more correct as it bypasses the equality operator and prevents use on structs.  I'd like to followup with another PR after this to replace the remaining == null with is null / for consistency.

iamcarbon:
Notable, they'll likely be at least one remaining round to enable the next batch of files. After that, we should be able to enable nullable at the project level, and remove the inline #nullable enables.

xoofx:
Thanks!

PR #525, opened 2021-03-10
Enable nullability, round 3 of x

iamcarbon:
Another big chunk of nullability annotations

### DIFF
--- a/src/Markdig/Renderers/TextRendererBase.cs
+++ b/src/Markdig/Renderers/TextRendererBase.cs
@@ -31,6 +31,7 @@ namespace Markdig.Renderers
         {
             if (writer is null) ThrowHelper.ArgumentNullException_writer();
             this.writer = writer;
+            this.writer.NewLine = "\n";
         }
 
         /// <summary>
--- a/src/Markdig/Renderers/TextRendererBase.cs
+++ b/src/Markdig/Renderers/TextRendererBase.cs
@@ -2,6 +2,8 @@
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
 
+#nullable enable
+
 using System;
 using System.Collections.Generic;
 using System.IO;
@@ -27,8 +29,8 @@ namespace Markdig.Renderers
         /// <exception cref="ArgumentNullException"></exception>
         protected TextRendererBase(TextWriter writer)
         {
-            if (writer == null) ThrowHelper.ArgumentNullException_writer();
-            Writer = writer;
+            if (writer is null) ThrowHelper.ArgumentNullException_writer();
+            this.writer = writer;
         }
 
         /// <summary>
@@ -72,13 +74,14 @@ namespace Markdig.Renderers
     {
         private class Indent
         {
-            private readonly string _constant;
-            private readonly Queue<string> _lineSpecific;
+            private readonly string? _constant;
+            private readonly Queue<string>? _lineSpecific;
 
             internal Indent(string constant)
             {
                 _constant = constant;
             }
+
             internal Indent(IEnumerable<string> lineSpecific)
             {
                 _lineSpecific = new Queue<string>(lineSpecific);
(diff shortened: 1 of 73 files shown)


iamcarbon:
Ready for review.

src/Markdig/Renderers/TextRendererBase.cs
```diff
@@ -27,8 +29,9 @@ public abstract class TextRendererBase : RendererBase
         /// <exception cref="ArgumentNullException"></exception>
         protected TextRendererBase(TextWriter writer)
         {
-            if (writer == null) ThrowHelper.ArgumentNullException_writer();
-            Writer = writer;
+            if (writer is null) ThrowHelper.ArgumentNullException_writer();
+            this.writer = writer;
+            this.writer.NewLine = "\n";
```
MihaZupan, line 34:
```suggestion
            Writer = writer;
```
We don't gain anything by duplicating this

src/Markdig/Syntax/ContainerBlock.cs
```diff
@@ -234,14 +236,14 @@ public void RemoveAt(int index)
...
-            Array.Sort(children, 0, Count, comparer);
+            if (comparer is null) ThrowHelper.ArgumentNullException(nameof(comparer));
+            Array.Sort(children, 0, Count, comparer!);
         }
 
         public void Sort(Comparison<Block> comparison)
         {
             if (comparison == null) ThrowHelper.ArgumentNullException(nameof(comparison));
-            Array.Sort(children, 0, Count, new BlockComparer(comparison));
+            Array.Sort(children, 0, Count, new BlockComparer(comparison)!);
```
MihaZupan, line 246: This should be needed right after the `new`

src/Markdig/Syntax/ContainerBlock.cs
```diff
@@ -305,9 +306,9 @@ public BlockComparer(Comparison<Block> comparison)
                 this.comparison = comparison;
             }
 
-            public int Compare(Block x, Block y)
+            public int Compare(Block? x, Block? y)
             {
-                return comparison(x, y);
+                return comparison(x!, y!);
             }
```
MihaZupan, line 312: This should either accept non-nullable blocks, or the comparer should allow nullable

src/Markdig/Renderers/TextRendererBase.cs
iamcarbon: Sadly, the analyzer doesn't follow the setter through and mark the writer field as non-null if we do this. We could set a MemberNotNull attribute and throw when setting the writer, but we also double up some logic if we do this.

src/Markdig/Syntax/ContainerBlock.cs
iamcarbon: This is an annoying case where the nullability contract of IComparer (below) forces us to define these as nullable.

``` int Compare([AllowNull] T x, [AllowNull] T y); ``` 

The System.Comparison delegate, sadly, does NOT define the system contract with nullable attributes. Our options are limited here since we don't control either implementation / contract.

iamcarbon: UPDATE: we can make the comparison explicitly accept nullable. But this forces us to suppress its use in Array.Sort -- and we don't actually have a case to compare against null. IComparer, regardless, requires us to accept null -- so I believe this is a valid case for suppression. We could also make this class oblivious.

src/Markdig/Syntax/ContainerBlock.cs
iamcarbon: I backed up out making the comparison delegate nullable, so these suppressions are no longer needed.

src/Markdig/Renderers/TextRendererBase.cs
MihaZupan: Unfortunate :/
This would be a fine place for `MemberNotNull`, but since we'd have to add a dummy class just for this, it's fine as is.

MihaZupan (approved):
Looks good, thanks!

xoofx:
Thanks!

PR #526, opened 2021-03-11
Complete nullable annotations

iamcarbon:
This PR completes the nullability annotations and enables nullable at the project level. 

There are 4 possibility breaking changes introduced to enforce initialization of non-nullable properties to the following classes.

HeadingLinkReferenceDefinition
FootnoteLink
FootnoteLinkReference
HtmlInline

While these classes have to be constructed with a non-null object, they remain fully mutable after construction to allow the DOM to be manipulated.

### DIFF
--- a/src/Markdig/Helpers/StringSlice.cs
+++ b/src/Markdig/Helpers/StringSlice.cs
@@ -83,7 +83,7 @@ namespace Markdig.Helpers
         /// <summary>
         /// The text of this slice.
         /// </summary>
-        public string Text;
+        public readonly string Text;
 
         /// <summary>
         /// Gets or sets the start position within <see cref="Text"/>.
--- a/src/Markdig/Helpers/StringSlice.cs
+++ b/src/Markdig/Helpers/StringSlice.cs
@@ -2,6 +2,8 @@
 // This file is licensed under the BSD-Clause 2 license. 
 // See the license.txt file in the project root for more information.
 
+#nullable disable
+
 using System;
 using System.Runtime.CompilerServices;
 
@@ -81,7 +83,7 @@ namespace Markdig.Helpers
         /// <summary>
         /// The text of this slice.
         /// </summary>
-        public readonly string Text;
+        public string Text;
 
         /// <summary>
         /// Gets or sets the start position within <see cref="Text"/>.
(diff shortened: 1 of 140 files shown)


src/Markdig/Helpers/StringSlice.cs
```diff
@@ -81,7 +83,7 @@ public StringSlice(string text, int start, int end, NewLine newLine)
         /// <summary>
         /// The text of this slice.
         /// </summary>
-        public readonly string Text;
+        public string Text;
```
MihaZupan, line 86: Why this change?

src/Markdig/Renderers/Normalize/Inlines/LinkInlineRenderer.cs
```diff
@@ -44,7 +44,7 @@ protected override void Write(NormalizeRenderer renderer, LinkInline link)
                 {
                     renderer.Write('(').Write(link.Url);
 
-                    if (!string.IsNullOrEmpty(link.Title))
+                    if (link.Title is { Length: > 0 })
```
MihaZupan, line 47: We don't *have* to change everything to pattern matching. Last time I checked it still had worse code generation, so as long as we don't do it on hot paths is should be fine.

MihaZupan (commented):
Aside from the StringSlice change, LGTM, thanks

src/Markdig/Helpers/StringSlice.cs
iamcarbon: Good catch! Thank you. This was accidental when backing out of changing this to a readonly property.

src/Markdig/Renderers/Normalize/Inlines/LinkInlineRenderer.cs
iamcarbon: Agreed -- and have seen similar reports of worse codegen for this pattern. This is another one-off to ensure for the analyzer to flow through the non-null check.

iamcarbon:
@MihaZupan -- Thank you for reviewing and catching that unintended change. Going to squeeze in one final commit to remove the existing #nullable enable directives and reduce churn.... and this should be GTG.

iamcarbon:
And done.

xoofx:
Amazing work @iamcarbon! 👍

PR #527, opened 2021-03-11
A few nullable changes

MihaZupan:
cc: @iamcarbon

### DIFF
--- a/src/Markdig.Tests/TestReferralLinks.cs
+++ b/src/Markdig.Tests/TestReferralLinks.cs
@@ -37,20 +37,6 @@ namespace Markdig.Tests
             Assert.That(html, Contains.Substring($"rel=\"{expected}\""));
         }
 
-
-        [Test]
-        public void TestLinksWithNullRel()
-        {
-            var markdown = "[world](http://example.com)";
-
-            var pipeline = new MarkdownPipelineBuilder()
-                .UseReferralLinks(null)
-                .Build();
-            var html = Markdown.ToHtml(markdown, pipeline);
-
-            Assert.That(html, !Contains.Substring("rel="));
-        }
-
         [Test]
         [TestCase(new[] { "noopener" }, "noopener")]
         [TestCase(new[] { "nofollow" }, "nofollow")]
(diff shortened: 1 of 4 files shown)


src/Markdig.Tests/TestReferralLinks.cs
```diff
@@ -37,20 +37,6 @@ public void TestLinksWithCustomRel(string[] rels, string expected)
...
-            var markdown = "[world](http://example.com)";
-
-            var pipeline = new MarkdownPipelineBuilder()
-                .UseReferralLinks(null)
-                .Build();
-            var html = Markdown.ToHtml(markdown, pipeline);
-
-            Assert.That(html, !Contains.Substring("rel="));
-        }
-
```
MihaZupan, line 53: I don't know if I'm missing something, but I don't understand this test.

If you don't want the `rel` being output... don't add the extension?

src/Markdig/Renderers/Html/Inlines/AutolinkInlineRenderer.cs
```diff
@@ -25,14 +25,22 @@ public bool AutoRelNoFollow
...
+                const string NoFollow = "nofollow";
+
+                if (value)
                 {
-                    Rel = string.IsNullOrEmpty(Rel) ? rel : Rel + $" {rel}";
+                    if (string.IsNullOrEmpty(Rel))
+                    {
+                        Rel = NoFollow;
+                    }
+                    else if (!Rel!.Contains(NoFollow))
```
MihaZupan, line 36: This `!` isn't needed on Core 3.1+, sadly older runtimes aren't smart enough to realize `string.IsNullOrEmpty` is a null check

iamcarbon: There's a NRE condition here that we're not handling correctly. We should rewrite this code to correctly handle a null Rel.

MihaZupan: It does handle it via `string.IsNullOrEmpty`

iamcarbon: Indeed. Missed the outer logic changes here. 👍

PR #528, opened 2021-03-11
Improve CQ1

iamcarbon:
This PR.

- Eliminates a few list allocations (and possible internal resizes) by using arrays
- Eliminates a queue allocation
- Ensures string comparisons are invariant
- Eliminates a possible NRE when allocating ReferralLinksExtension with a null value and subsequently adding additional rels


MihaZupan:
There's some overlap with #527 on the `Rel` changes.

I don't think we'd want to bump the target just for this. We did already but reverted that, see #466.

iamcarbon:
@MihaZupan  -- It looks like you beat me to helping fix these violations!

Let's get your change committed first and I'll rebase and start a new PR with a clean history.

PR #529, opened 2021-03-12
Cache the MarkdownParser & Processors

MihaZupan:
- `MarkdownParser` is made static (had private ctor before)
- `Block/InlineProcessor` got a set of Setup/Rent/Reset/Release methods

```c#
Markdown.Parse("Hello world!");
```

Before:
image

After:
image

### DIFF
--- a/src/Markdig/Parsers/BlockProcessor.cs
+++ b/src/Markdig/Parsers/BlockProcessor.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using Markdig.Helpers;
 using Markdig.Syntax;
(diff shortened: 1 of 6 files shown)


MihaZupan:
I think this makes of free of "Constant overhead" in terms of allocations on the `Parse`/`ToHtml` paths.

src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -16,165 +17,159 @@ namespace Markdig.Parsers
...
-            Open(document);
-        }
+        public bool TrackTrivia { get; private set; }
 
         public bool SkipFirstUnwindSpace { get; set; }
 
+        private bool ContinueProcessingLine { get; set; }
+
         /// <summary>
-        /// Gets the new blocks to push. A <see cref="BlockParser"/> is required to push new blocks that it creates to this property.
```
xoofx, line 66: Why are the order of the properties are all changed in this file? Could this be done in a separate PR?

MihaZupan: Oh yeah I can try to revert this part. I was doing it to triple-check all state was getting reset

MihaZupan: Changed back to original order

PR #530, opened 2021-03-12
Improve Code Quality

iamcarbon:
This PR:

- Compares strings using StringComparison.Ordinal or pattern matching to avoid culture differences
- Writes  chars, where possible, to avoid passing references
- Eliminates a few allocations
- Enables nullable in a few more places
- Ensures CodeInline.Content is initialized by adding a constructor
- Seals renderer classes to ensure they are not subclassed

### DIFF
--- a/src/Markdig/Parsers/Inlines/AutolinkInlineParser.cs
+++ b/src/Markdig/Parsers/Inlines/AutolinkInlineParser.cs
@@ -35,10 +35,9 @@ namespace Markdig.Parsers.Inlines
             int column;
             if (LinkHelper.TryParseAutolink(ref slice, out string? link, out bool isEmail))
             {
-                processor.Inline = new AutolinkInline()
+                processor.Inline = new AutolinkInline(url: link)
                 {
                     IsEmail = isEmail,
-                    Url = link,
                     Span = new SourceSpan(processor.GetSourcePosition(saved.Start, out line, out column), processor.GetSourcePosition(slice.Start - 1)),
                     Line = line,
                     Column = column
--- a/src/Markdig/Parsers/Inlines/AutolinkInlineParser.cs
+++ b/src/Markdig/Parsers/Inlines/AutolinkInlineParser.cs
@@ -35,7 +35,7 @@ namespace Markdig.Parsers.Inlines
             int column;
             if (LinkHelper.TryParseAutolink(ref slice, out string? link, out bool isEmail))
             {
-                processor.Inline = new AutolinkInline(url: link)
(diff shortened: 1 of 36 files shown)


iamcarbon:
Ready for review.

src/Markdig/Parsers/Inlines/AutolinkInlineParser.cs
```diff
@@ -35,10 +35,9 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             int column;
             if (LinkHelper.TryParseAutolink(ref slice, out string? link, out bool isEmail))
             {
-                processor.Inline = new AutolinkInline()
+                processor.Inline = new AutolinkInline(url: link)
```
MihaZupan, line 38: Tiny nit: the parameter should be obvious here

MihaZupan (approved):
Looks good! I like that we're using the right `StartsWith` semantics now, really unfortunate the defaults are all mixed up for comparrisons now

iamcarbon:
Squeezed in one more change to seal the Renderer classes.

MihaZupan:
What's the aim behind sealing the renderers? I could see inheriting from them being useful

iamcarbon:
I can see the usefulness of extending the syntax nodes when introducing custom extensions, but not for the renders whom's sole responsibility is to implement a single Write method). Let me know if you you can think of a useful case to inherit these.

By sealing these now, we can make future changes knowing that they weren't subclassed. Sealing a class may also enable the JIT to devirtualize certain calls (avoiding the method lookup table).

Happy to backout, regardless, if you think they should remain unsealed.

xoofx:
> I can see the usefulness of extending the syntax nodes when introducing custom extensions, but not for the renders whom's sole responsibility is to implement a single Write method. Is there a specific use case for inheriting the a Render class that you can think of?

The use case is that the write method is overridable: If you want to wrap an existing renderer, you can simply inherit from it, override the write method and emit some text before/after calling the base method.

> By sealing these now, we can make future changes knowing that they weren't subclassed. 

This behavior has been around for 5 years now. Making them sealed could break lots of users, so it's not really something that we should do. 😉  We have cases where we broke some code (e.g typo renaming...etc.) in the past but overall, when there are overridable methods and that's ok for a user to override it, I don't like the idea to restrict this scenario. I'm always super frustrated myself when using a library sealing all its types that I could see a benefit of overriding.

iamcarbon:
@xoofx 👍

I gone ahead and reverted these changes and put and end to my sealing crusade.

xoofx:
Thanks!

PR #531, opened 2021-03-14
A handful of codegen improvements

MihaZupan:
- Added a `PeekChar()` overload of `PeekChar(int)` to `ICharIterator`. This method was already present on `StringSlice`, but since it was missing on the interface, callers were often binding to the slower overload
- Added `SkipChar` to `ICharIterator`
- Replaced `CurrentCharacter == '\0'` and `Length == 0` with faster `IsEmpty` checks
- Avoided a bunch of struct copies in `StringLineGroup`
- Changed the `NewLine.Length()` to a bit operation
- Simplified/optimized logic in `HtmlHelper`
- Avoided the `ToUtf32` iterator allocations in Globalization extension


iamcarbon:
Consider renaming SkipChar to Advance() and adding an Advance(count) overload to match SequenceReader naming.

PR #548, opened 2021-05-03
JiraLinkInlineParser.cs - Make digits in JiraKey's posible

Mysteryduck001:
A digit can also be part of the project key for example T3ST is a valid project key and T3ST-123 is thus a valid Jira issue number but it will not be recognized by this tool. 
I propose to add an OR IsDigit() check like: 'while (current.IsAlphaUpper() || current.IsDigit())'
Thanks for considering to update.

### DIFF
--- a/src/Markdig/Extensions/JiraLinks/JiraLinkInlineParser.cs
+++ b/src/Markdig/Extensions/JiraLinks/JiraLinkInlineParser.cs
@@ -40,7 +40,13 @@ namespace Markdig.Extensions.JiraLinks
             var startKey = slice.Start;
             var endKey = slice.Start;
 
-            //read as many uppercase characters as required - project key
+            // the first character of the key can not be a digit.
+            if (current.IsDigit())
+            {
+                return false;
+            }
+
+            // read as many uppercase characters or digits as required - project key
             while (current.IsAlphaUpper() || current.IsDigit())
             {
                 endKey = slice.Start;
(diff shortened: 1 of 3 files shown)


src/Markdig/Extensions/JiraLinks/JiraLinkInlineParser.cs
```diff
@@ -41,7 +41,7 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             var endKey = slice.Start;
 
             //read as many uppercase characters as required - project key
```
MihaZupan, line 43:
```suggestion
            // read as many uppercase characters or digits as required - project key
```

MihaZupan (approved):
Thanks! Could you please also add a test for this case?

Adding an example here: https://github.com/xoofx/markdig/blob/master/src/Markdig.Tests/Specs/JiraLinks.md
And then re-running the `SpecFileGen` project

src/Markdig/Extensions/JiraLinks/JiraLinkInlineParser.cs
Mysteryduck001: Hi, I see everything is added for the test en verified. 
I've added a test to check if the first char in the key is not a digit as this is not allowed by Jira.

xoofx:
Thanks!

PR #578, opened 2021-12-03
ADD: Context to Renderer

grishat:
Added MarkdownParserContext to Renderer.

I wrote my own LinkExtension render `TryLinkInlineRenderer`, and need context access from `HtmlRenderer`, that available only in parser.
```
    public void Setup(MarkdownPipeline pipeline, IMarkdownRenderer renderer)
    {
        if (renderer is HtmlRenderer htmlRenderer)
        {
            var inlineRenderer = htmlRenderer.ObjectRenderers.FindExact<LinkInlineRenderer>();
            if (inlineRenderer != null)
            {
                inlineRenderer.TryWriters.Remove(TryLinkInlineRenderer);
                inlineRenderer.TryWriters.Add(TryLinkInlineRenderer);
            }
        }
    }

    private bool TryLinkInlineRenderer(HtmlRenderer renderer, LinkInline link)
    {
        ...
       var propValue = (string?)renderer.Context?.Properties["PropName"];
        ...
    }
```

I think the context should be available both in the parser and in the render. This required minimal changes.
What do you think?

### DIFF
--- a/src/Markdig/Markdown.cs
+++ b/src/Markdig/Markdown.cs
@@ -72,7 +72,7 @@ namespace Markdig
 
             var document = MarkdownParser.Parse(markdown, pipeline, context);
 
-            var renderer = new NormalizeRenderer(writer, options);
+            var renderer = new NormalizeRenderer(writer, options, context);
             pipeline.Setup(renderer);
 
             renderer.Render(document);
@@ -97,7 +97,7 @@ namespace Markdig
 
             var document = MarkdownParser.Parse(markdown, pipeline, context);
 
-            return ToHtml(document, pipeline);
+            return ToHtml(document, pipeline, context);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


src/Markdig/Markdown.cs
```diff
@@ -107,13 +107,13 @@ public static string ToHtml(string markdown, MarkdownPipeline? pipeline = null,
...
         /// <exception cref="ArgumentNullException">if markdown document variable is null</exception>
-        public static string ToHtml(this MarkdownDocument document, MarkdownPipeline? pipeline = null)
+        public static string ToHtml(this MarkdownDocument document, MarkdownPipeline? pipeline = null, MarkdownParserContext? context = null)
         {
             if (document is null) ThrowHelper.ArgumentNullException(nameof(document));
 
             pipeline ??= DefaultPipeline;
 
-            using var rentedRenderer = pipeline.RentHtmlRenderer();
+            using var rentedRenderer = pipeline.RentHtmlRenderer(null, context);
```
MihaZupan, line 116:
```suggestion
            using var rentedRenderer = pipeline.RentHtmlRenderer(writer: null, context);
```

src/Markdig/MarkdownPipeline.cs
```diff
@@ -100,17 +100,19 @@ internal sealed class HtmlRendererCache : ObjectCache<HtmlRenderer>
 
             private readonly MarkdownPipeline _pipeline;
             private readonly bool _customWriter;
+            private readonly MarkdownParserContext? _context;
 
-            public HtmlRendererCache(MarkdownPipeline pipeline, bool customWriter = false)
+            public HtmlRendererCache(MarkdownPipeline pipeline, bool customWriter = false, MarkdownParserContext? context = null)
```
MihaZupan, line 105: The `MarkdownParserContext` may be different for each parsing operation and shouldn't be cached between uses of the renderer.

Instead of storing the context on `HtmlRendererCache`, set it after renting it like we [do with the writer](https://github.com/xoofx/markdig/blob/bd124f073472f4d394091da3fdea179253183f39/src/Markdig/MarkdownPipeline.cs#L87-L90) and then null it out in `Reset`.

MihaZupan (commented):
Seems reasonable to me to extend the context to rendering

grishat:
I get you, it turned out to be a little more complicated than I thought.
I found walkaround for me. I think I need to close this PR. I can't run the tests either on my local machine.

PR #586, opened 2022-01-19
.NET 6.0

tibel:
use .NET 6.0 SDK and update test and benchmark dependencies

### DIFF
--- a/src/Markdig.Tests/Markdig.Tests.csproj
+++ b/src/Markdig.Tests/Markdig.Tests.csproj
@@ -1,7 +1,7 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
     <OutputType>Exe</OutputType>
     <IsPackable>false</IsPackable>
     <StartupObject>Markdig.Tests.Program</StartupObject>
--- a/src/Markdig.Tests/Markdig.Tests.csproj
+++ b/src/Markdig.Tests/Markdig.Tests.csproj
@@ -8,13 +8,13 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="coverlet.msbuild" Version="2.9.0">
+    <PackageReference Include="coverlet.msbuild" Version="3.1.0">
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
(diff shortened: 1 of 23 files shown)


src/UnicodeNormDApp/Program.cs
```diff
@@ -11,10 +11,10 @@ class Program
...
         static void Main(string[] args)
         {
-
-            var httpClient = new WebClient();
-
-            var data = httpClient.DownloadString("http://www.unicode.org/Public/UCD/latest/ucd/NormalizationTest.txt");
+            using var httpClient = new HttpClient();
+            using var response = httpClient.GetAsync("http://www.unicode.org/Public/UCD/latest/ucd/NormalizationTest.txt").Result;
+            using var content = response.Content;
+            var data = content.ReadAsStringAsync().Result;
```
MihaZupan, line 17:
```suggestion
            var data = await new HttpClient().GetStringAsync("http://www.unicode.org/Public/UCD/latest/ucd/NormalizationTest.txt");
```
We can just make `Main` `async` here.

src/Markdig.Tests/Markdig.Tests.csproj
```diff
@@ -1,20 +1,20 @@
 <Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>net5.0</TargetFramework>
+    <TargetFramework>net6.0</TargetFramework>
```
MihaZupan, line 4: I was under the impression we had multi-targeted tests as well.
If we haven't done so before I guess keeping just 6.0 here is fine.

MihaZupan (approved):

tibel:
Added another commit to fix build warnings (mostly xml doc).

src/global.json
```diff
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "5.0.100",
+    "version": "6.0.x",
```
xoofx, line 3: In the past, this was not working and required an explicit version (not like in global.json). Are you sure that it doesn't fetch the latest?

tibel: I'm not aware of this issue. Do you have a link to check if this is fixed already?
Alternatively I can change this to an explicit version, if you want.

xoofx, line 3:
```suggestion
    "version": "6.0.100",
```

xoofx: I don't think it is fixed as the [documentation](https://docs.microsoft.com/en-us/dotnet/core/tools/global-json?tabs=netcore3x#version) still states that you need an actual version and wildcards (e.g even `x`) are not supported

So just to be safe, easier to put an actual version. I pushed the suggestion.

xoofx:
Thanks!

PR #588, opened 2022-01-20
Diagrams are not correctly escaped

tibel:
The issue with diagrams extension is that not all characters are escaped and thus invalid HTML could be generated.
A diagram sample is https://www.nomnoml.com/ that contains `<table>` in the diagram text.
Currently markdig is not escaping `>` but is escaping `<`.
This results in parsing errors in mermaid and nomnoml.

`CodeBlockRenderer` uses `softEsacape=false` and with `BlocksAsDiv` it uses `softEscape=true`.
According to documentation `BlocksAsDiv` should not change escaping behavior.

Please let me know when there are questions or something is missing.


tibel:
Sorry, this was my fault.
You can ignore this.

PR #590, opened 2022-01-21
Fix link reference definition parse bug with title and CRLF

yufeih:
Fix LinkReferenceDefinition parse bug when CRLF is used as line ending: https://github.com/dotnet/docfx/issues/7850


xoofx:
Thanks!

PR #611, opened 2022-03-20
Improve rendering performance

### LINKED ISSUE #547: Unsafe caching of a previous renderer
There seems to be an issue in the way how a previous renderer is cached in [RendererBase.Write](https://github.com/xoofx/markdig/blob/0a0040450fd4f774dacc71bcaf0ed1651a792cea/src/Markdig/Renderers/RendererBase.cs#L134-L158) method.

Each registered renderer has an `Accept` method saying whether the renderer can process a particular object. Since the `Accept` method is virtual, it's possible to create custom selective renderers. The problem with a cached renderer is that it's never tested whether it can accept another object or not - the implementation makes an assumption just by the object type, whereas a custom `Accept` method can have much more criteria to check.

Moreover, even if the `Accept` method had invoked on the cached previous render before using it, we'd have still had an issue, because some renderer(s) located in the pipeline prior to the cached one could accept a new object while rejecting the previous object of the same type. That contemplation leads us to the conclusion that previous renderer cache will always be unsafe and risky to use.

Here's an issue we ran into. We needed to emit custom markup for HTML anchors defined as `[label](#anchor)`. Those anchors are actually regular `LinkInline` objects. We wrote a custom renderer with an overridden `Accept` method checking that a link URL is an anchor URL, so it's only invoked for anchors. The `AnchorRenderer` was placed before `LinkInlineRenderer` in the pipeline, so the `AnchorRenderer` was supposed to be invoked for each link parsed from a markdown document. However, if a paragraph has two URLs: a regular one, and an anchor - the first URL is processed with `LinkInlineRenderer` that is cached and then reused for the anchor. This way the `AnchorRenderer` is never actually invoked.

The workaround is to create a parser that would produce an object of different type, smth like `AnchorLinkInline` so it's not picked up by `LinkInlineRenderer`. Which is fine, but requires more entities implemented and adds extra computation costs on parsing stage.

MihaZupan:
Closes #547

API change as mentioned in #547:
```diff
public interface IMarkdownObjectRenderer
{
-   bool Accept(RendererBase renderer, MarkdownObject obj);
+   bool Accept(RendererBase renderer, Type objectType);
}
```

Before
|           Method |     SourceText |      Mean |    Error |   StdDev |    Median |   Gen 0 | Allocated |
|----------------- |--------------- |----------:|---------:|---------:|----------:|--------:|----------:|
| ToHtmlRenderOnly |  MarkdigReadme | 127.31 us | 0.296 us | 0.433 us | 127.36 us |  8.5449 |     36 KB |
| ToHtmlRenderOnly | TracingArticle |  19.92 us | 0.156 us | 0.209 us |  20.04 us |  1.3428 |      6 KB |

After
|           Method |     SourceText |       Mean |     Error |    StdDev |   Gen 0 | Allocated |
|----------------- |--------------- |-----------:|----------:|----------:|--------:|----------:|
| ToHtmlRenderOnly |  MarkdigReadme |  69.793 us | 0.0402 us | 0.0550 us |  8.6670 |     36 KB |
| ToHtmlRenderOnly | TracingArticle |   8.990 us | 0.0297 us | 0.0426 us |  1.3428 |      6 KB |


src/Markdig/Renderers/RendererBase.cs
```diff
@@ -16,21 +16,34 @@ namespace Markdig.Renderers
...
+                var renderer = ObjectRenderers[i];
+                if (renderer.Accept(this, objectType))
+                {
+                    _renderersPerType[typeHandle] = renderer;
+                    return renderer;
+                }
+            }
+
+            _renderersPerType[typeHandle] = null;
+            return null;
```
MihaZupan, line 43: We weren't marking "unknown" types before, meaning that if a type didn't have an associated renderer, we would loop through the list every time.

I.e. for every `LinkReferenceDefinition`, `ContainerInline`, `ContainerBlock`, etc. we would fallback from `Dictionary` to looping through all renderers.

I guess is that this is why the `Type.IsAssignableFrom` was so much more expensive compared to `is`.

PR #622, opened 2022-04-15
Vectorize CharacterMap.IndexOfOpeningCharacter

MihaZupan:
A 1-2.5% improvement for parsing (on x86 LE hardware only for now)

For long runs without an opening character, the vectorized loop is ~8x faster.

| Method | AdvancedPipeline |     Mean |   Error |  StdDev |   Median |
|------- |----------------- |---------:|--------:|--------:|---------:|
|  Parse |            False | 262.6 us | 0.73 us | 3.64 us | 263.5 us |
|  Parse |             True | 366.9 us | 0.43 us | 2.16 us | 367.2 us |

| Method | AdvancedPipeline |     Mean |   Error |  StdDev |
|------- |----------------- |---------:|--------:|--------:|
|  Parse |            False | 259.5 us | 0.18 us | 0.91 us |
|  Parse |             True | 357.9 us | 0.41 us | 2.06 us |

### DIFF
--- a/src/Markdig/Helpers/CharacterMap.cs
+++ b/src/Markdig/Helpers/CharacterMap.cs
@@ -7,6 +7,11 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Runtime.CompilerServices;
+#if NETCOREAPP3_1_OR_GREATER
+using System.Numerics;
+using System.Runtime.Intrinsics;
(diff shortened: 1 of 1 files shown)


xoofx:
Nice!

MihaZupan:
Some good news, this now turned into a discussion about where we can improve the existing helper methods in .NET: https://github.com/dotnet/runtime/issues/68328

PR #627, opened 2022-04-20
Update commonmark spec 0.30

### LINKED ISSUE #563: CommonMark spec version 0.30 was released
0.30 was released on 19th of June: https://spec.commonmark.org/0.30/ ([diff](https://spec.commonmark.org/0.30/changes.html))

I haven't yet taken the time to see if there are any interesting changes, but opening an issue here to raise awareness.

xoofx:
Fixes #563

Still an issue with the [Example 539](https://spec.commonmark.org/0.30/#example-539)

```md
[ẞ]

[SS]: /url
```
 
It seems that compare `OrdinalIgnoreCase` doesn't support Unicode Case folding? Or this particular character: [ẞ Latin Capital Letter Sharp S](https://unicode-table.com/en/1E9E/).

@MihaZupan thoughts?

### DIFF
--- a/src/Markdig.Tests/Specs/CommonMark.generated.cs
+++ b/src/Markdig.Tests/Specs/CommonMark.generated.cs
@@ -14,8 +14,8 @@ namespace Markdig.Tests.Specs.CommonMarkV_0_29
         // ---
         // title: CommonMark Spec
         // author: John MacFarlane
-        // version: 0.29
-        // date: '2019-04-06'
+        // version: '0.30'
+        // date: '2021-06-19'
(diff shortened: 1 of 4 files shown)


MihaZupan:
Opened https://github.com/dotnet/runtime/issues/68287 to see if .NET can do this out of the box or if we'd have to roll our own logic here

xoofx:
> That API does not exist on net452 though... so time to remove that target (again 😄)?

Thank you! The problem is small enough that we can ifdef it for now. net452 will not be completely 0.30 compliant, but it is so minor that it's probably ok.

PR #632, opened 2022-04-22
Don't try to use the MT pointer on unsupported platforms

### LINKED ISSUE #631: WebAssembly functionally broken between 0.28.1 and 0.29.0
I was testing with some very simple code and I have observed the following:  

This code does work in a .Net 6.0 Console application with version 0.29.0 & 0.28.1
This code does work in a .Net 6.0 Blazor Web Assembly project with version 0.28.1 and several earlier versions  (28.0, 27.0, 22.0).
This code does not work in a .Net 6.0 Blazor Web Assembly project with version 0.29.0 and later.

By 'not work' what I observered is that the result of .ToHtml is an empty string. It is not null. No exceptions are thrown, 

                .UseAdvancedExtensions()
                .Build();
                var md = @"# Header";
                var result = Markdown.ToHtml(md, pipeline);
                Console.WriteLine(result);
                return result;

So as far as I can tell, something between 28.1 an 29.0 broke the ability for the library to function correctly within Blazor WebAssembly projects. I did review https://github.com/xoofx/markdig/compare/0.28.1...0.29.0 but nothing there jumped out at me as an obvious culprit.

Unfortunately its not that easy to debug .Net code running as WASM in a browser, so I am running up against my ability to work out what is going on here. I did attempt to set the Builder's DebugLog to Console.Out, to see if that would get me anything, but then a code search has turned up that almost nothing writes to DebugLog.

At the moment, I will be forced to stay on 0.28.1.

MihaZupan:
Fixes #631

### DIFF
--- a/src/Markdig/Renderers/RendererBase.cs
+++ b/src/Markdig/Renderers/RendererBase.cs
@@ -4,7 +4,9 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
 using Markdig.Helpers;
 using Markdig.Syntax;
 using Markdig.Syntax.Inlines;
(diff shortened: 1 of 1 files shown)


xoofx:
Hey @MihaZupan,

So I merged your PR, ran a benchmark with it rendering the `spec.md` and reverted to using a plain `Type` for the dictionary key and compared, and on my machine, I don't see a difference in the results... so I have pushed the commit 94e07d1 to revert that optim. It is simpler to maintain and safer, hope you won't mind!

MihaZupan:
Don't mind at all, this is definitely the most "insane" of the optimizations :).

I'm surprised you didn't see a perf difference. I'll have to rerun the numbers again.

I was seeing
`MT ptr` is faster than `Type.GetTypeHandle` is faster than `Type`.
`RuntimeTypeHandle` is at least safe to use from a maintainability perspective.

Between `MT` and the type handle, I was seeing a few % of time spent in `GetTypeHandle` in profiles that prompted me to look for an alternative.

xoofx:
> MT ptr is faster than Type.GetTypeHandle is faster than Type.

But the code still has to perform a `obj.GetType()` to [pass it to the renderer](https://github.com/xoofx/markdig/blob/f2ca6be7a62e33ba7dee728ff1696056c71b9f84/src/Markdig/Renderers/RendererBase.cs#L36). I pushed a new commit to f09d030 to pass only the `objectType` directly to `GetRendererInstance` so that it should avoid the double call to `GetType()`. Otherwise, I don't see why exactly a `Type.GetTypeHandle` + `obj.GetType()` would be faster than a plain `obj.GetType()`. The dictionary is going to take the `GetHashCode` on the type so it might be more costly (because it has to go through the object header to freeze it), but compare to the rest of the code involved, I'm still skeptical on the impact.

MihaZupan:
> But the code still has to perform a obj.GetType() to [pass it to the renderer](https://github.com/xoofx/markdig/blob/f2ca6be7a62e33ba7dee728ff1696056c71b9f84/src/Markdig/Renderers/RendererBase.cs#L36).
> I don't see why exactly a Type.GetTypeHandle + obj.GetType() would be faster than a plain obj.GetType().

`GetRendererInstance` is called once per object type, the result is then cached. Its cost in the benchmark is effectively 0.

The only cost that really matters is the cost to:
- Compute the key - `GetType` vs `Type.GetTypeHandle`
- Dictionary lookup

On the fast path, all you are doing is a single `Type.GetTypeHandle` for the dictionary lookup.
The `GetType` you need for the renderer is only called once per type in `GetRendererInstance`.

MihaZupan:
> ran a benchmark with it rendering the spec.md

Just so I can augment my future benchmarks, by `spec.md` do you mean the whole CommonMark spec file?

xoofx:
> Just so I can augment my future benchmarks, by spec.md do you mean the whole CommonMark spec file?

Yeah, but it was the full `Markdown.ToHtml`, not just the renderer. If I rerun things with only the renderer:

1. `GetType()`: 550.3 us
2. `Type.GetTypeHandle()`: 500.3 us
3. `Unsafe Method table`: 478.4 us 

So 1. vs 3. is about 14% faster, but 2. is already giving the majority of the boost.
If I run a full `ToHtml`, it goes around `2.1ms`, so in that case, the performance benefit is quite less impactful.

So we can keep `Type.GetTypeHandle()`?

MihaZupan:
> So we can keep Type.GetTypeHandle()?

Yeah, that would be my suggestion.

This is the kind of numbers I was seeing for just rendering the Markdig readme (in this case even skipping the final string allocation):
- `GetType`: 73.27 us
- `Type.GetTypeHandle`: 62.20 us

PR #637, opened 2022-05-14
Reduce the overhead of Parse calls

MihaZupan:
Saves 8 % (~65 ns) on my machine when parsing `Markdown.Parse("foo", _pipeline)` with 
```c#
new MarkdownPipelineBuilder().UsePreciseSourceLocation();
```

and 15 % (~150 ns) with `UseAdvancedExtensions`.


xoofx:
Thanks!

PR #644, opened 2022-06-08
added support for ToPlainText for pipetables extension

PaulVrugt:
This PR adds support for `ToPlainText` for the pipetables extension. Before this, it would simply return all the table html elements, and render plaintext within the table. With this PR, all the table elements are removed from the plaintext output as you would expect as a result of `ToPlainText`


xoofx:
Thanks1

PR #649, opened 2022-07-17
Align Whitespace and Punctuation definitions with CommonMark

### LINKED ISSUE #645: Converting emphasis with angled quotation marks
Hi,

I'm trying to convert a document which contains `"«_word_»"` string. As you can see on example below, the parser cannot recognize it as emphasis:

``` csharp
var html1 = Markdown.ToHtml("«_word_»"); // "<p>«_word_»</p>\n"
```

But `"_«word»_"` has been converted ok:

``` csharp
var html2 = Markdown.ToHtml("_«word»_"); // "<p><em>«word»</em></p>\n"
```

I'm using Markdig 0.30.2. Is it a bug? If yes, is there any workaround to avoid the issue? Thanks.

MihaZupan:
Fixes #645

### DIFF
--- a/src/Markdig.Tests/TestCharHelper.cs
+++ b/src/Markdig.Tests/TestCharHelper.cs
@@ -1,0 +1,92 @@
+using System.Collections.Generic;
+using System.Globalization;
+using Markdig.Helpers;
+using NUnit.Framework;
+
+namespace Markdig.Tests
+{
+    public class TestCharHelper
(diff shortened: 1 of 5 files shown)


xoofx:
PR failing because of coverage upload, will need to investigate this, sorry for the trouble.

xoofx:
Thanks for the fix!

Issue #671, opened 2022-10-13
MediaLinkExtensions: Controls


LukeTOBrien:
Hello,

I am creating a Static Website Generator and as one of my test I was creating a site with a video background, so this video has attributes `muted`, `loop` and `autoplay` but I do not want the `controls` attribute.  
In [MediaLinkExtension.cs line 113](https://github.com/xoofx/markdig/blob/98c687b4ed8b7a7004837948fe18dcb8013528f7/src/Markdig/Extensions/MediaLinks/MediaLinkExtension.cs#L113) there is code that adds the `controls` attribute but they is no way to say that I do not want this attribute to be added.  
Do you think a boolean flag could be added to `MediaOptions` so that I can spefify that the attribute should not be added?

```cs
if (Options.Controls)
{
    htmlAttributes.AddPropertyIfNotExist("controls", null);
}
```

Thanks,

Luke

xoofx:
> Do you think a boolean flag could be added to MediaOptions so that I can spefify that the attribute should not be added?

Yeah, completely, PR welcome.

PR #672, opened 2022-10-13
Option to set controls attribute for MediaLinks

LukeTOBrien:
Relating to issue https://github.com/xoofx/markdig/issues/671

### DIFF
--- a/src/Markdig/Extensions/MediaLinks/MediaOptions.cs
+++ b/src/Markdig/Extensions/MediaLinks/MediaOptions.cs
@@ -16,7 +16,7 @@ namespace Markdig.Extensions.MediaLinks
         {
             Width = "500";
             Height = "281";
-            Controls = true;
+            AddControlsProperty = true;
             Class = "";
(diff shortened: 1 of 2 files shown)


src/Markdig/Extensions/MediaLinks/MediaOptions.cs
```diff
@@ -90,6 +91,8 @@ public MediaOptions()
 
         public string Height { get; set; }
 
+        public bool Controls { get; set; }
```
MihaZupan, line 94: How about something more explicit, e.g. `AddControlsProperty`?

LukeTOBrien: Well you could make the same argument about `width` and `height`, both of them are optional.  
`Controls` is the name of the HTML attribute and I think `Controls` true or false will be easily understood, no need to be explicit

LukeTOBrien: @MihaZupan  I don't think it's needed myself but I have renamed the property name as you suggested

PR #676, opened 2022-10-25
Use Span.Fill instead of for loop in ValueStringBuilder.Append

KeterSCP:
This PR replaces manual for loop in ValueStringBuilder.Append with Span.Fill, which is much faster on big buffers due to vectorization.

### DIFF
--- a/src/Markdig/Helpers/ValueStringBuilder.cs
+++ b/src/Markdig/Helpers/ValueStringBuilder.cs
@@ -87,11 +87,7 @@ namespace Markdig.Helpers
                 Grow(count);
             }
 
-            Span<char> dst = _chars.Slice(_pos, count);
-            for (int i = 0; i < dst.Length; i++)
-            {
-                dst[i] = c;
-            }
+            _chars.Slice(_pos, count).Fill(c);
             _pos += count;
         }
 


MihaZupan:
> which is much faster on big buffers due to vectorization

You are right in general, but the two usages I see in YARP will call it with:
- https://github.com/xoofx/markdig/blob/936fe35460a2fe645b41aff55fb1d9581fbce552/src/Markdig/Parsers/Inlines/CodeInlineParser.cs#L88 `closeTicks` which is `1` 99% of the time
- https://github.com/xoofx/markdig/blob/936fe35460a2fe645b41aff55fb1d9581fbce552/src/Markdig/Syntax/LeafBlock.cs#L92 `AddTab` which is always `[1, 4]`

So for these uses, the change would actually be a regression compared to for loop

KeterSCP:
> So for these uses, the change would actually be a regression compared to for loop

Yep, should have checked actual usages first. Thanks for noticing @MihaZupan. If there are no plans for using this Append overload with big enough buffers, I can close the PR

xoofx:
Thanks for the PR! Though, I agree with @MihaZupan. Also a performance improvement should be better measured on the wall time of Markdig on a typical document, so that we can better assess if it's worth it or not.

PR #678, opened 2022-11-25
Copyedit the grid table extension documentation

waldyrious:
Various small tweaks for syntax consistency, readability, clarity and correctness.

### DIFF
--- a/src/Markdig.Tests/Specs/GridTableSpecs.md
+++ b/src/Markdig.Tests/Specs/GridTableSpecs.md
@@ -4,7 +4,8 @@ This section describes the different extensions supported:
 
 ## Grid Table
 
-A grid table allows to have multiple lines per cells and allows to span cells over multiple columns. The following shows a simple grid table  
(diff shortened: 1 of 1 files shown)


src/Markdig.Tests/Specs/GridTableSpecs.md
```diff
@@ -4,9 +4,10 @@ This section describes the different extensions supported:
 
 ## Grid Table
 
-A grid table allows to have multiple lines per cells and allows to span cells over multiple columns. The following shows a simple grid table  
+A grid table allows having multiple lines per cells and allows spanning cells over multiple columns.
+The following shows a simple grid table:
 
-```
+```````````````````````````````` example
```
waldyrious, line 10: I suspect this change may not be desirable; the other example blocks below appear to have both the source code and the rendered HTML. I suppose they're used for tests?

waldyrious: Yeah, clearly this should not be an `example` block. I've reverted this change.

xoofx:
Thanks!

PR #685, opened 2022-12-17
Drop netcoreapp3.1 target

iamcarbon:
This PR drops support for the netcoreapp3.1 target (now out of support).

.netcoreapp3.1 will get the netstandard2.1 target after this update.


omer-iqbal:
Hi,

It might also help readers if the readme.md at root is updated which lists the compatible frameworks.

Thx

iamcarbon:
@omer-iqbal The library continues to support ALL supported .NET frameworks -- and this PR does not change that. The netstandard2.1 target allows the library to continue to be used by netcoreapp3.1.

Nuget does a great job at showing the libraries compatibility: https://www.nuget.org/packages/Markdig

xoofx:
Thanks!

Issue #684, opened 2022-12-11
Adding IDs to footnote renderer


omer-iqbal:
I am working with documents and Islamic commentaries. My goal is to preserve the original footnote numbers or labels so they can be searched and found by the users and so users can also cite the original footnote when needed. So I need to preserve the original footnote.

There are two issues:
1) For larger works, like books and commentaries, the original text is broken into multiple sections (which can be hundreds or even thousands for multi-volume works) and each section will be stored in its own markdown.
2) Some works have footnotes that use non-numeric labels. For example, [this Quran commentary book](https://www.cpsglobal.org/books/Quran-Commentary-English.pdf) (e.g. page 112) uses letters *a*, *b* etc. for footnotes.

Based on that, it looks like footnotes need a feature where the original ID can be preserved in the markup instead of using the Order that is currently used.

If this proposal seems acceptable, I can try to make a pull request (not sure though what the time frame might look like).

Right now, I am thinking that I will add another property (say, Id) to [FootnoteGroup](https://github.com/xoofx/markdig/blob/master/src/Markdig/Extensions/Footnotes/FootnoteGroup.cs). Based on an option provided, [HtmlFootnoteGroupRenderer](https://github.com/xoofx/markdig/blob/master/src/Markdig/Extensions/Footnotes/HtmlFootnoteGroupRenderer.cs) will either use Order or this new Id both for text and, potentially, for use in the hrefs.

Comments/suggestions/feedback welcome.

I looked at [this list of issues](https://github.com/xoofx/markdig/issues?q=footnote) and didn't find a similar issue.

xoofx:
> Based on that, it looks like footnotes need a feature where [Footnote.Label](https://github.com/xoofx/markdig/blob/master/src/Markdig/Extensions/Footnotes/Footnote.cs#L25) can be used in the markup instead of the [Footnote.Order]
(https://github.com/xoofx/markdig/blob/master/src/Markdig/Extensions/Footnotes/Footnote.cs#L30) property.

I'm confused, because Footnote in markdig do support using label instead of order [here in the specs](https://github.com/xoofx/markdig/blob/master/src/Markdig.Tests/Specs/FootnotesSpecs.md), so why this is not working for your case?

PR #686, opened 2022-12-17
Add support for preserving original label in footnotes & omit back links (fixes issue #684)

omer-iqbal:
- Added support for preserving original labels in footnotes
- Added an option to omit the backlink in footnotes
- As a side effect of generating specs, comments in GridTableSpec are also updated

This will resolve issue # 684.


omer-iqbal:
Need to add one more commit to this PR

PR #687, opened 2022-12-17
Add support for preserving original label in footnotes & omit back links (fixes issue #684)

omer-iqbal:
1.  Added support for preserving original labels in the footnotes
2.  Added an option to not add back links in the footnotes
3.  As a side effect of generating specs, comments in GridTableSpec are also updated

This will resolve issue #684.


omer-iqbal:
Need to change some additional layout issues

PR #693, opened 2023-01-19
Enable ImplictUsings and use file-scoped namespaces

iamcarbon:
This PR is up for feedback on whether we should enable file-scoped namespaces on some of all of the markdig projects.

Pulling the indentation in 4 lines on the tests is a fairly nice improvement for readability -- particularly on a 15" laptop.


iamcarbon:
@MihaZupan @xoofx Let me know what you think. If we do make this change on the main library -- it would make the change history harder to follow -- but improve the experience working with those files moving forward.

xoofx:
> @MihaZupan @xoofx Let me know what you think. If we do make this change on the main library -- it would make the change history harder to follow -- but improve the experience working with those files moving forward.

Yeah, please go ahead in applying this to the rest of the library, it should be ok. Thanks!

MihaZupan:
I like the feature, thanks!
My only concern is the quality of the automatic refactoring - when we did this in YARP it messed up indentation in random places. We should just double-check that doesn't happen before merging.

iamcarbon:
@MihaZupan Agreed on the caution -- I used the refactoring tools to do this file by file, checking them as I went along to make sure everything was indented correctly. We should be OK!

iamcarbon:
@xoofx @MihaZupan Ready for review. I also fixed a couple spelling errors in the comments while I was double checking the indentation levels.

iamcarbon:
Here's a direct link to review ignoring whitespace changes: https://github.com/xoofx/markdig/pull/693/files?diff=unified&w=1

xoofx:
Thanks a lot!

PR #696, opened 2023-02-11
Fix incorrect setext heading source span

artempyanykh:
Previously, heading's span would be 3-6; where 6 is outside the bounds of the original string.

For the context, the original bug report in Marksman [#156](https://github.com/artempyanykh/marksman/issues/156) (I use markdig there).


artempyanykh:
Haha, no, not really 🙂 Thanks for catching this @MihaZupan!

xoofx:
Thanks!

Issue #715, opened 2023-06-13
MediaLinks Extension doesn't work on relative links


RickStrahl:
It appears that the `.UseMediaLinks()` extension does not detect media when the URL is not an http link, but rather a relative path.

The following demonstrates:

![image](https://github.com/xoofx/markdig/assets/1374013/d14e2695-94f8-423f-a343-638dc2afd31d)

Note that the only media link created correctly is the one that has the full HTTP link while the relative link/no path ones do not render.

Here's a Gist with the above code:
https://gist.github.com/RickStrahl/36035c87787fc24e37861059e40a6f8b

RickStrahl:
For reference here's a screen shot of Markdown Monster rendering the above links as well as a relative image and an explicit HTML video tag:

![image](https://github.com/xoofx/markdig/assets/1374013/4ede3f7b-93be-4021-a55c-52d1658059ea)

xoofx:
Medialinks extension only supports [absolute URL](https://github.com/xoofx/markdig/blob/b7cb169fd3a924e1e0d4e2e487a22c51caaedd10/src/Markdig/Extensions/MediaLinks/MediaLinkExtension.cs#L54-L67). It could support relative with an option to allow them.
I don't have time to implement such feature but PR welcome if someone is looking for this.

RickStrahl:
Do you think this needs an option? Seems like this should just work otherwise it's confusing why some media links work and some do not. 

Looks like the only other parser that actually supports this is PanDoc, but it does turn relative paths into respective `<video>` and `<audio>` links along with fully qualified Urls.

If I find some time I'll take a look.

xoofx:
> Do you think relative paths need an option? Seems like this should just work otherwise it's confusing why some media links work and some do not.

Problem is more backward compatibility that could start to break existing code that were relying on this behavior.

PR #723, opened 2023-07-11
Add relative path support for Audio and Video Urls to  MediaLinks Extension

RickStrahl:
Allow detect and embed relative URLs for image tag usage of audio and video links. Relates to #715.

Updated linking looks like this:

image

Note relative links now work picking up media files out of the current folder as well as online links.


RickStrahl:
LOL - sure why not?

PR #726, opened 2023-07-25
Make AllowNull internal to prevent conflicts

mkapahnke:
- Unity uses a version of "netstandard2.1", but already includes the AllowNull attribute
- This will prevent conflicts with other packages, that might include this attribute as well


mkapahnke:
I guess, that this should not have an affect on other code.

PR #759, opened 2023-12-11
Make StringLineGroup returns a count limited Enumerator

### LINKED ISSUE #757: Apply `foreach` on `class StringLineGroup` got phantom empty line, is it expected?
I' m writing code to dump a `FencedCodeBlock`, here is the function:
```csharp
        private static void HandleCodeBlock(FencedCodeBlock codeblock, int level)
        {
            Console.ForegroundColor = ConsoleColor.DarkYellow;
            if (!string.IsNullOrEmpty(codeblock.Info))
            {
                PrintIndent(level);
                Console.WriteLine($"lang: {codeblock.Info}, info args: {codeblock.Arguments}");
            }
            Console.ForegroundColor = ConsoleColor.White;

            foreach (var line in codeblock.Lines)
                Console.WriteLine(line);
        }
```
Then I see output contains suspicious empty line like this:
```plaintext
  FencedCodeBlock: 36/692-735 @1
 characters.sifeng.voice("c1s1", 3)


```
And the corresponding markdown is:
\`\`\`
characters.sifeng.voice("c1s1", 3)
\`\`\`

After some investigation, I found the enumerator of `StringLineGroup` returns internal array's enumerator directly. The array contains reserved slot for future item adding, that's the reason why we see phantom empty lines.

Shall we change the `GetEnumerator()` for `StringLineGroup`? I can provide some help.

Akarinnnnn:
Fixes #757, before we return the array enumerator directly, enumerate it will get phantom empty lines.

### DIFF
--- a/src/Markdig/Helpers/StringLineGroup.cs
+++ b/src/Markdig/Helpers/StringLineGroup.cs
@@ -218,7 +218,7 @@ public struct StringLineGroup : IEnumerable
 
     IEnumerator IEnumerable.GetEnumerator() 
     {
-        (IEnumerator)GetEnumerator();
+        return GetEnumerator();
     }
 
     private void IncreaseCapacity()
--- a/src/Markdig/Helpers/StringLineGroup.cs
+++ b/src/Markdig/Helpers/StringLineGroup.cs
@@ -187,7 +187,7 @@ public struct StringLineGroup : IEnumerable
         }
     }
 
-    private struct Enumerator : IEnumerator
+    public struct Enumerator : IEnumerator
     {
         private readonly StringLineGroup _parent;
         private int _index;
@@ -210,10 +210,15 @@ public struct StringLineGroup : IEnumerable
             _index = -1;
         }
     }
-
-    IEnumerator IEnumerable.GetEnumerator()
+    
+    public Enumerator GetEnumerator()
     {
         return new Enumerator(this);
+    }
+
+    IEnumerator IEnumerable.GetEnumerator() 
+    {
+        (IEnumerator)GetEnumerator();
     }
 
     private void IncreaseCapacity()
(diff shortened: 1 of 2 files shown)


src/Markdig/Helpers/StringLineGroup.cs
```diff
@@ -187,9 +187,40 @@ public void Trim()
...
+        public bool MoveNext()
+        {
+            if (++_index < _parent.Count)
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
```
xoofx, line 212:
```suggestion
            return ++_index < _parent.Count;
```

src/Markdig/Helpers/StringLineGroup.cs
```diff
@@ -187,9 +187,33 @@ public void Trim()
         }
     }
 
+    private struct Enumerator : IEnumerator
```
xoofx, line 190: Could you add a `public Enumerator GetEnumerator()` to avoid the boxing of this struct?

Akarinnnnn: Do you mean turn explicit implementation of `IEnumerator.GetEnumerator()` into `Enumerator GetEnumerator()`? I guess it's not intended for directly use since it's explicit impl.

xoofx: No. It's called (enumerator) duck-typing. You keep the explicit implementation but you declare a public method with the same signature but that returns the struct `Enumerator GetEnumerator()`.

When using foreach, it will detect this method and avoid calling the explicit implementation.

Akarinnnnn: Added.

xoofx:
Thanks!

PR #760, opened 2023-12-14
Fix source span calculation for LineBreakInline

zickb:
Fix the source span calculation of the last `LineBreakInline` in paragraph when multiple blocks exists.

I have added the source span calculation for the last `LineBreakInline` in https://github.com/xoofx/markdig/pull/733, but the logic does not handle multiple blocks correctly. This corrects the logic.


xoofx:
Thanks!

PR #761, opened 2023-12-14
Use C# 12 syntax

iamcarbon:
Posting for early feedback on the introduction / use of collection expressions & primary expressions.


xoofx:
Thanks!

PR #769, opened 2024-02-09
Eliminate various string allocations

iamcarbon:
This PR introduces an overload on TextRendererBase that writes a character a specific number of times, and utilizes it to eliminate various string allocations.


MihaZupan (approved):
Thanks.

I wouldn't be surprised if there's more low-hanging fruit around the normalize/roundtrip renderers, I never really looked into them.

xoofx:
Thanks!

PR #776, opened 2024-03-04
Add support for GitHub alert blocks

### LINKED ISSUE #774: Add support for GitHub Markdown Alerts
See https://github.com/orgs/community/discussions/16925

```md
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.
```

rendered as:

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

> [!TIP]
> Optional information to help a user be more successful.

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

> [!CAUTION]
> Negative potential consequences of an action.

xoofx:
Fixes #774

This extension does not support roundtrip.


xoofx:
@MihaZupan I added you as a reviewer. First time in years that I haven't implemented an extension to markdig 😅

src/Markdig/Extensions/Alerts/AlertBlock.cs
```diff
@@ -0,0 +1,33 @@
...
+/// <summary>
+/// A block representing an alert quote block.
+/// </summary>
+public class AlertBlock : QuoteBlock
+{
+    /// <summary>
+    /// Creates a new instance of this block.
+    /// </summary>
+    /// <param name="kind"></param>
+    public AlertBlock(StringSlice kind) : base(null!)
```
gfoidl, line 19: Instead of the `null!` should https://github.com/xoofx/markdig/blob/fcb56fb03701cc1a07a52769908579adb8927ee2/src/Markdig/Syntax/QuoteBlock.cs#L22 be updated to allow `null`?

Especially as https://github.com/xoofx/markdig/blob/fcb56fb03701cc1a07a52769908579adb8927ee2/src/Markdig/Syntax/Block.cs#L23 allows `null`.

src/Markdig/Extensions/Alerts/AlertInlineParser.cs
```diff
@@ -0,0 +1,126 @@
...
+
+            c = slice.NextChar();
+        }
+
+        var alertBlock = new AlertBlock(alertType)
+        {
+            Span = new SourceSpan(startPosition, processor.GetSourcePosition(slice.Start - 1)),
+            TriviaSpaceAfterKind = new StringSlice(slice.Text, start, end),
+            Line = line,
+            Column = column
```
MihaZupan, line 105:
```suggestion
            Span = quoteBlock.Span,
            TriviaSpaceAfterKind = new StringSlice(slice.Text, start, end),
            Line = quoteBlock.Line,
            Column = quoteBlock.Column
```

I think we might want to inherit the positions of the quote block here, otherwise the first `> ` is skipped.

MihaZupan (approved):
Left a couple notes, but this looks good to me.

It feels a bit odd that we're hardcoding SVGs, but I don't know of a good alternative. Users can customize the logic themselves anyway.

jorge-moreira:
Hi Alexandre,

I was trying your code and I found a situation where the `AlertBlock` might not be parsed.

For this example, I added a code block inside the alert:
```
> [!NOTE]
> Highlights information that users should take into account, even when skimming.
> Testing rendering for multiple lines
> ```csharp
> var test = "I can also add code to panels
> ```
> `Inline code testing` 
```

This is the result on GitHub markdown:
> [!NOTE]
> Highlights information that users should take into account, even when skimming.
> Testing rendering for multiple lines
> ```csharp
> var test = "I can also add code to panels
> ```
> `Inline code testing` 

Although, it will never be rendered, since is not parsed as an `AlertBlock`.

If you need more info regarding this, I'm available to help.

xoofx:
> I was trying your code and I found a situation where the AlertBlock might not be parsed.

Thanks, good catch, I have a fix. I mistakenly forgot that inline parser happens after all blocks have been parsed.

src/Markdig/Extensions/Alerts/AlertInlineParser.cs
```diff
@@ -0,0 +1,126 @@
...
+        };
+
+        alertBlock.GetAttributes().AddClass("markdown-alert");
+        alertBlock.GetAttributes().AddClass($"markdown-alert-{alertType.ToString().ToLowerInvariant()}");
+
+        // Replace the quote block with the alert block
+        var parentQuoteBlock = quoteBlock.Parent!;
+        var indexOfQuoteBlock = parentQuoteBlock.IndexOf(quoteBlock);
+        parentQuoteBlock.RemoveAt(indexOfQuoteBlock);
+        parentQuoteBlock.Insert(indexOfQuoteBlock, alertBlock);
```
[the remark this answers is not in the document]
xoofx: Fixed via commit af38833

src/Markdig/Extensions/Alerts/AlertInlineParser.cs
```diff
@@ -0,0 +1,126 @@
...
+
+        while (quoteBlock.Count > 0)
+        {
+            var block = quoteBlock[0];
+            quoteBlock.RemoveAt(0);
+            alertBlock.Add(block);
+        }
```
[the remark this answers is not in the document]
xoofx: Fixed via commit af38833

src/Markdig/Extensions/Alerts/AlertInlineParser.cs
xoofx: Fixed via commit af38833

src/Markdig/Extensions/Alerts/AlertInlineParser.cs
```diff
@@ -0,0 +1,126 @@
...
+    {
+        OpeningCharacters = ['['];
+    }
+
+    public override bool Match(InlineProcessor processor, ref StringSlice slice)
+    {
+        // We expect the alert to be the first child of a quote block. Example:
+        // > [!NOTE]
+        // > This is a note
+        if (!(processor.Block is ParagraphBlock paragraphBlock) || !(paragraphBlock.Parent is QuoteBlock quoteBlock) || quoteBlock.Count != 1 || paragraphBlock.Inline?.FirstChild != null)
```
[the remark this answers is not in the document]
xoofx: Fixed via commit af38833

src/Markdig/Extensions/Alerts/AlertBlock.cs
xoofx: Fixed by commit 64dc4e6

xoofx:
Thanks all for the review!

jorge-moreira:
Hi @xoofx 

I was trying the new version and it seems that the paragraphs within the `AlertBlock` are not being rendered.

Markdown example:
```
> [!NOTE]
> Highlights information that users should take into account, even when skimming.
> 
> Testing rendering for multiple lines
> 
> `Inline code testing`
> 
> Other line
> 
> > Nested quote
> >
> > Final nested quote line
> 
> Final line of alert
```

Expected:
> [!NOTE]
> Highlights information that users should take into account, even when skimming.
> 
> Testing rendering for multiple lines
> 
> `Inline code testing`
> 
> Other line
> 
> > Nested quote
> >
> > Final nested quote line
> 
> Final line of alert

HTML rendered result:
```
<div class="markdown-alert markdown-alert-note">
<p class="markdown-alert-title"><svg viewBox="0 0 16 16" version="1.1" width="16" height="16" aria-hidden="true"><path d="M0 8a8 8 0 1 1 16 0A8 8 0 0 1 0 8Zm8-6.5a6.5 6.5 0 1 0 0 13 6.5 6.5 0 0 0 0-13ZM6.5 7.75A.75.75 0 0 1 7.25 7h1a.75.75 0 0 1 .75.75v2.75h.25a.75.75 0 0 1 0 1.5h-2a.75.75 0 0 1 0-1.5h.25v-2h-.25a.75.75 0 0 1-.75-.75ZM8 6a1 1 0 1 1 0-2 1 1 0 0 1 0 2Z"></path></svg>Note</p>
<p>Highlights information that users should take into account, even when skimming.</p>
<p></p>
<p></p>
<p></p>
<blockquote>
<p></p>
<p></p>
</blockquote>
<p></p>
</div>
```

Do you want me to open an issue with this info?

xoofx:
> Do you want me to open an issue with this info?

Yes please.

PR #782, opened 2024-03-15
Add bootstrap alert renderer

### LINKED ISSUE #781: Feature-request: Use bootstrap alerts for alert blocks
In case bootstrap is used, alert-blocks should utilize the corresponding CSS classes.

Examples:

```html
<div class="alert alert-primary" role="alert">
  A simple primary alert—check it out!
</div>
<div class="alert alert-secondary" role="alert">
  A simple secondary alert—check it out!
</div>
<div class="alert alert-success" role="alert">
  A simple success alert—check it out!
</div>
<div class="alert alert-danger" role="alert">
  A simple danger alert—check it out!
</div>
<div class="alert alert-warning" role="alert">
  A simple warning alert—check it out!
</div>
<div class="alert alert-info" role="alert">
  A simple info alert—check it out!
</div>
<div class="alert alert-light" role="alert">
  A simple light alert—check it out!
</div>
<div class="alert alert-dark" role="alert">
  A simple dark alert—check it out!
</div>
```

Will look like this:

image

Or using the dark theme:

image

---

Thank you for considering the inclusion of this feature. I'll be happy to help.

Abrynos:
Thank you for considering the inclusion of this merge request.

Closes #781 

For testing I used the following MD:

```markdown
# TODO

> [!NOTE]
> This is the newly generated NOTE block
> 
> And this is the second paragraph

> [!TIP]
> This is the newly generated TIP block
>
> And this is the second paragraph

> [!IMPORTANT]
> This is the newly generated IMPORTANT block
>
> And this is the second paragraph

> [!WARNING]
> This is the newly generated WARNING block
>
> And this is the second paragraph

> [!CAUTION]
> This is the newly generated CAUTION block
> 
> And this is the second paragraph

```

To produce this HTML:

```html
<h1 id="todo">TODO</h1>
<div class="markdown-alert markdown-alert-note alert alert-primary" role="alert">
<p>This is the newly generated NOTE block</p>
<p style="margin-bottom: 0">And this is the second paragraph</p>
</div>
<div class="markdown-alert markdown-alert-tip alert alert-success" role="alert">
<p>This is the newly generated TIP block</p>
<p style="margin-bottom: 0">And this is the second paragraph</p>
</div>
<div class="markdown-alert markdown-alert-important alert alert-info" role="alert">
<p>This is the newly generated IMPORTANT block</p>
<p style="margin-bottom: 0">And this is the second paragraph</p>
</div>
<div class="markdown-alert markdown-alert-warning alert alert-warning" role="alert">
<p>This is the newly generated WARNING block</p>
<p style="margin-bottom: 0">And this is the second paragraph</p>
</div>
<div class="markdown-alert markdown-alert-caution alert alert-danger" role="alert">
<p>This is the newly generated CAUTION block</p>
<p style="margin-bottom: 0">And this is the second paragraph</p>
</div>
```

Which (with my HTML header and footer) looks like this:

image


Abrynos:
Oh and of course a screenshot with the light theme of bootstrap:

image

src/Markdig/Extensions/Bootstrap/BootstrapAlertRenderer.cs
```diff
@@ -0,0 +1,72 @@
...
+using Markdig.Renderers.Html;
+using Markdig.Syntax;
+
+namespace Markdig.Extensions.Bootstrap;
+
+/// <summary>
+/// A HTML renderer for a <see cref="AlertBlock"/> that adds bootstrap classes.
+/// </summary>
+/// <seealso cref="HtmlObjectRenderer{AlertBlock}" />
+public class BootstrapAlertRenderer : HtmlObjectRenderer<AlertBlock>
```
xoofx, line 17: I would think that you would not need to create a new `BootstrapAlertRenderer` class.

You can add the attributes as we do directly in the boostrap extension and then you can add if not already the existing [AlertBlockRenderer](https://github.com/xoofx/markdig/blob/master/src/Markdig/Extensions/Alerts/AlertBlockRenderer.cs) and set the RenderKind property to discard the rendering of the SVG. In the end it should be equivalent to the code you wrote with a bit less duplication.

Also, in general, a renderer should not modify the model that it renders (so AddProperty should not be done there)

xoofx: Thinking about it, I would not remove the Note text with the SVG from the Alert block when activating Bootstrap. We should probably only keep bootstrap extension to add the attributes.

Abrynos: > Thinking about it, I would not remove the Note text with the SVG from the Alert block when activating Bootstrap. [...]

I disagree for multiple reasons:
- I believe that this library should not modify WHAT is rendered (the user determines that by writing markdown), but only HOW the content is rendered. Here we are already using bootstrap (at the users choice), which makes an effort to highlight this visually already. Therefore we don't need extra text as we do without bootstrap.
- That text has nothing to distinguish it from the content. It looks like part of it rather than some sort of headline that was added automatically.
- The icons are quite hard to make out when using the bootstrap dark theme.
- Not utilizing boostrap-icons but additional SVGs. However, we should not blindly use boostrap-icons either because that is a separate CSS/font that is not shipped with bootstrap-core by default.
- There is literally no space between the icon and the text, which makes it look crowded.

image

xoofx: Fair enough

xoofx:
Thanks!

MihaZupan:
As a follow-up, I realize that GH allows the `NOTE` / `TIP`, etc. in any casing, not just all upper.
We might want to match that in our Alert block logic.

Abrynos:
> As a follow-up, I realize that GH allows the `NOTE` / `TIP`, etc. in any casing, not just all upper. We might want to match that in our Alert block logic.

Good catch! Should I make another PR, @xoofx ?

xoofx:
> > As a follow-up, I realize that GH allows the `NOTE` / `TIP`, etc. in any casing, not just all upper. We might want to match that in our Alert block logic.
> 
> Good catch! Should I make another PR, @xoofx ?

Sure, go ahead. 🙂

PR #784, opened 2024-03-18
Make alert block headers case-invariant

Abrynos:
This is a follow-up for #782

### DIFF
--- a/src/Markdig/Extensions/Alerts/AlertBlockRenderer.cs
+++ b/src/Markdig/Extensions/Alerts/AlertBlockRenderer.cs
@@ -61,7 +61,9 @@ public class AlertBlockRenderer : HtmlObjectRenderer<AlertBlock>
     /// <param name="kind">The kind of the alert to render</param>
     public static void DefaultRenderKind(HtmlRenderer renderer, StringSlice kind)
     {
-        string? html = kind.AsSpan() switch
+        Span<char> upperKind = stackalloc char[kind.Length];
+        kind.AsSpan().ToUpperInvariant(upperKind);
+        string? html = upperKind switch
         {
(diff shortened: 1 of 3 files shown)


src/Markdig/Extensions/Alerts/AlertBlockRenderer.cs
```diff
@@ -61,7 +61,9 @@ protected override void Write(HtmlRenderer renderer, AlertBlock obj)
     /// <param name="kind">The kind of the alert to render</param>
     public static void DefaultRenderKind(HtmlRenderer renderer, StringSlice kind)
     {
-        string? html = kind.AsSpan() switch
+        Span<char> upperKind = stackalloc char[kind.Length];
```
xoofx, line 64: This method might be better protected by checking e.g `kind.Length >= 16` and exit before trying to stackalloc.

src/Markdig/Extensions/Bootstrap/BootstrapExtension.cs
```diff
@@ -61,7 +61,10 @@ private static void PipelineOnDocumentProcessed(MarkdownDocument document)
                     var attributes = node.GetAttributes();
                     attributes.AddClass("alert");
                     attributes.AddProperty("role", "alert");
-                    string? @class = alertBlock.Kind.AsSpan() switch
+
+                    Span<char> upperKind = new char[alertBlock.Kind.Length];
```
xoofx, line 65: Similarly, we can use stackalloc here and protect it under a check `kind.Length >= 16`

Abrynos: I didn't use `stackallock` here for a reason. It is only freed after the method returns so it might lead to a temporary memory leak in case of huge documents with many alert blocks.

MihaZupan: You could switch on `alertBlock.Kind.ToUpperInvariant()` instead (avoids the allocation if it's already upper case).
Or move the `stackalloc char[16]` out of the loop

Abrynos:
I've also just noticed that unknown alert types blend into the background without a border:
image
image

xoofx:
Thanks!

PR #785, opened 2024-03-20
Fix issues for math span calculation

toothache:
A follow up fix for #779 . Math span end position should also consider the situation when the math inline is closed with whitespace.

### DIFF
--- a/src/Markdig.Tests/TestSourcePosition.cs
+++ b/src/Markdig.Tests/TestSourcePosition.cs
@@ -698,7 +698,7 @@ literal      ( 0, 2)  2-3
     [Test]
     public void TestMathematicsInline()
     {
-        //     01 23456789AB
+        //     01 23456789ABCDEF
         Check("0\n012 $abcd$ 321", @"
 paragraph    ( 0, 0)  0-15
 literal      ( 0, 0)  0-0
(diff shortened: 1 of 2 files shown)


toothache:
> I'm not sure what exactly the fix for set text heading is trying to address. The following [seems to be correct](https://babelmark.github.io/?text=Hello%0A%3D%3D%3D+%0A%0ATest%0A%3D%3D%3Dn%0A) for CommonMark implementations, including markdig.

markdig is parsing the second set text heading as heading. But it shouldn't, correct?

Results 1 implementation#4
[markdig](https://github.com/lunet-io/markdig/) 0.35.0.0C# 

<h1>
  Hello
</h1>
<h1>
  Test
</h1>

xoofx:
> markdig is parsing the second set text heading with an extra character as heading. But it shouldn't, correct?

Sorry, completely misread babelmark results. You are absolutely correct.

src/Markdig/Parsers/ParagraphBlockParser.cs
```diff
@@ -176,6 +176,11 @@ private static char GetHeadingChar(ref StringSlice line, ref int count)
                 return headingChar;
             }
 
+            if (!line.CurrentChar.IsSpaceOrTab())
```
xoofx, line 179: I fixed it with 000393f46a3e8ed4bac5a685fdafa48c0dfb1a21, could you remove this change from this PR? Thanks!

xoofx:
Thanks!

toothache:
> Thanks!

Hi, @xoofx . Is it possible to release a new version containing the fix? Thanks!

PR #802, opened 2024-06-09
Add ConfigureHtmlRenderer to MarkdownPipelineBuilder

yngvedh:
I needed to set up image url rewriting in a project of mine and found it a bit tedious not having a way to set it up in the pipeline. So I added the ability to configure the `HtmlRenderer` from the pipeline builder. As an added bonus, this works with the renderer cache as well.

I did not add any tests but updated the BaseUrl test to configure the html renderer using the pipeline builder.
I did not update changelog.md since it already has been neglected for some time.
I have not spent any time figuring out whether any of the documentation should be updated, please advise.
I could not find any existing PR or issue on this exact issue so I took the liberty to raise this draft PR.

UPDATE:
I added a bare bones builder for HtmlRenderer letting me set the one field I need as a POC. It is basically a factory for `TextRendererBase` which I also refactored the cache mechanism to use so that too should work for any text renderer. This goes for the `ToHtml` method since it uses the caching mechanism to get hold of its renderer.

### DIFF
--- a/src/Markdig.Tests/RoundtripSpecs/Inlines/TestNullCharacterInline.cs
+++ b/src/Markdig.Tests/RoundtripSpecs/Inlines/TestNullCharacterInline.cs
@@ -22,16 +22,15 @@ namespace Markdig.Tests.RoundtripSpecs.Inlines
         // do not unintentionally use the expected parameter
         private static void RoundTrip(string markdown, string expected)
         {
-            var pipelineBuilder = new MarkdownPipelineBuilder();
-            pipelineBuilder.EnableTrackTrivia();
-            MarkdownPipeline pipeline = pipelineBuilder.Build();
+            var pipeline = new MarkdownPipelineBuilder()
+                .EnableTrackTrivia()
+                .ConfigureRoundtripRenderer()
+                .Build();
+
             MarkdownDocument markdownDocument = Markdown.Parse(markdown, pipeline);
-            var sw = new StringWriter();
-            var rr = new RoundtripRenderer(sw);
+            var result = Markdown.ToHtml(markdownDocument, pipeline);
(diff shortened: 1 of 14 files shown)


xoofx:
Thanks. It looks ok at a quick glance.

yngvedh:
Great, I'll see if I can find the time to flesh it out with more properties and renderers.

yngvedh:
So I added builders for all kinds of renderers along with all fields. I also added doc comments and updated some tests.

Now, I'm not 100% sure all those fields should be available through the builder. In particular the `EnableHtmlFor*` options because they seem to only be used internally by the different renderers/extensions. Please let me know whether I should drop some of them.

I'm also not 100% sure all those renderers should be available through the builder at either so let me know if any of them shouldn't and I'll have them removed.

MihaZupan:
A couple observations:
- HtmlRenderer is by far the most used thing. I think it'd be fine to scope any such helpers to just that.
- I find it odd that we'd effectively duplicate all properties on renderers with fluent extension methods.
  - Could we expose a single `ConfigureHtmlRenderer(this builder, Action<HtmlRenderer>)` instead?
- Having the configuration only apply to renderers created internally by the cache is odd / inconsistent with other configurations on the pipeline builder.
  - It's likely you could implement this as an extension instead, modifying the renderer as part of the extension's renderer configuration. That'd mean this configuration would also apply to custom renderers created via the pipeline.

xoofx:
Indeed, seeing more changes coming, I agree the points from @MihaZupan

Mazahuacar (approved):

MihaZupan:
@yngvedh are you still interested in working on this one? (totally fine if not)

PR #826, opened 2024-11-21
Fix extra line feeds in link title

snnz:
Extra line feeds appear in the parsed link titles containing Windows-style line breaks, because \n that is peeked after \r is not skipped.


xoofx:
Thanks! Good catch, could you add a test?

snnz:
I've added a test. In the current version, it fails with the message:
_Expected string length 16 but was 17. Strings differ at index 9.   Expected: "this\ris\r\na\ntitle"   But was:  "this\ris\r\n\na\ntitle"_

PR #828, opened 2024-11-30
Add .NET 9 target, drop .NET 6

MihaZupan:
Unlike in the past, there is now a reason to bump TFMs even when not explicitly using new APIs due to language changes that can target better overloads when recompiling (e.g. `params span` or `OverloadResolutionPriority`).
I don't know of any places where that would kick in for Markdig right now, but there are some neat new APIs in 9.0 that might be interesting to play with here.

I've also added back older TFMs to the test project so we're not completely without coverage (given we have a bit of conditionally-compiled code & polyfills). Skipped framework though cuz CI was being weird 🤷‍♂️

Some initial numbers without any Markdig changes on a ___large___ document (.NET 9 perf blog post ~570 KB):

| Method         | Runtime  | Mean       | Error    |
|--------------- |--------- |-----------:|---------:|
| Parse          | .NET 8.0 | 1,015.0 us |  3.78 us |
| Parse          | .NET 9.0 |   926.8 us |  3.58 us |
|                |          |            |          |
| ParseAdvanced  | .NET 8.0 | 2,892.6 us | 17.57 us |
| ParseAdvanced  | .NET 9.0 | 2,803.9 us |  4.20 us |
|                |          |            |          |
| Render         | .NET 8.0 |   459.2 us |  5.76 us |
| Render         | .NET 9.0 |   425.7 us |  5.38 us |
|                |          |            |          |
| RenderAdvanced | .NET 8.0 |   585.2 us | 10.13 us |
| RenderAdvanced | .NET 9.0 |   489.8 us |  5.35 us |

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,6 @@ jobs:
   build:
     uses: xoofx/.github/.github/workflows/dotnet.yml@main
     with:
-      dotnet-version: '6.0 8.0'
+      dotnet-version: '6.0 8.0 9.0'
     secrets:
       NUGET_TOKEN: ${{ secrets.NUGET_TOKEN }}    
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,6 +13,9 @@ jobs:
   build:
     uses: xoofx/.github/.github/workflows/dotnet.yml@main
     with:
-      dotnet-version: '6.0 8.0 9.0'
+      dotnet-version: |
+        6.0
+        8.0
+        9.0
     secrets:
       NUGET_TOKEN: ${{ secrets.NUGET_TOKEN }}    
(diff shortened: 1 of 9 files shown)


.github/workflows/ci.yml
```diff
@@ -13,6 +13,9 @@ jobs:
   build:
     uses: xoofx/.github/.github/workflows/dotnet.yml@main
     with:
-      dotnet-version: '6.0 8.0'
```
MihaZupan, line 16: Passing `dotnet-version: '6.0 8.0 9.0'` to the setup-dotnet action ends up building the latest LTS + the first item in the list.
With `6.0 8.0` that just happens to match the intended result.

xoofx:
Thanks!

snnz:
This is nice. By those whose projects are targeting .NET6 will now receive a bonus with the updates, in a form of .NET Standard 2.1 DLL that misses some features of the .NET versions, like local inits switched off and everything that depends on the "NET" condition not used.

MihaZupan:
Such projects should likely focus on moving to a supported .NET runtime before updating their Markdig version then, especially if performance is a concern

snnz:
Well, that's unfortunate. Doesn't seem right somehow. An unsupported runtime is just a runtime that ceased to progress. The applications and libraries targeting it are not supposed to perform worse than they did before in subsequent runtimes.

PR #835, opened 2024-12-18
Include opening and closing pipes in the table span

### LINKED ISSUE #775: PipeTableParser strip opening and ending characters
[Here is my code](https://share.linqpad.net/ugvsf9ot.linq) that can run under LINQPad:

```cs
void Main()
{
	var builder = new MarkdownPipelineBuilder();
	
	var pt = new PipeTableExtension();
	pt.Setup(builder);

	var pipeline = builder.Build();
	
	var markdownText = GetText();

	var document = Markdown.Parse(markdownText, pipeline);

	var blocks = document.ToList();
	
	foreach (var item in blocks)
	{
		string blockText = ExtractText(markdownText, item);
		blockText.Pre().Dump("Parsed Output");
	}
}

public static string ExtractText(string text, Block item)
{
	var pos = item.ToPositionText(); // $295, 0, 13301-13375

	// parse "$295, 0, 13301-13375" into lines, columns, range variables
	var line = pos.Split(",")[0].TrimStart('$').Trim();
	var column = pos.Split(",")[1].Trim();

	var range = pos.Split(",")[2].Trim();
	var start = range.Split("-")[0];
	var end = range.Split("-")[1];

	var blockText = text.Substring(int.Parse(start), int.Parse(end) - int.Parse(start) + 1);
	return blockText;
}

string GetText() => """

| Attributes               | Details                                           |
|:---                      |:---                                               |
| `<docs-card-container>`  | All cards must be nested inside a container       |
| `title`                  | Card title                                        |
| card body contents       | Anything between `<docs-card>` and `</docs-card>` |
| `link`                   | (Optional) Call to Action link text               |
| `href`                   | (Optional) Call to Action link href               |

""";
```

When running this code, the parsed output will remove the opening and ending characters. It seems a bug.

image

snnz:
Fixes #775

### DIFF
--- a/src/Markdig.Tests/TestSourcePosition.cs
+++ b/src/Markdig.Tests/TestSourcePosition.cs
@@ -808,6 +808,29 @@ literal      ( 4, 2) 13-13
     }
 
     [Test]
+    public void TestPipeTable3()
+    {
+        //     01234 5678 9ABCD
+        Check("|a|b\n-|-\n0|1|\n", @"
+table        ( 0, 0)  0-12
+tablerow     ( 0, 1)  1-3
+tablecell    ( 0, 1)  1-1
+paragraph    ( 0, 1)  1-1
+literal      ( 0, 1)  1-1
+tablecell    ( 0, 3)  3-3
+paragraph    ( 0, 3)  3-3
+literal      ( 0, 3)  3-3
+tablerow     ( 2, 0)  9-11
+tablecell    ( 2, 0)  9-9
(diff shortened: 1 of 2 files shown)


xoofx:
Could you add a test for this one to validate the fix and prevent from a regression? If we want to rewrite the pipetable parser with a different (and simpler) strategy, more compatible with GitHub, it would help a lot.

snnz:
I made a test based on an existing one.

PR #836, opened 2024-12-18
Fix an error in the AbbreviationParser.

snnz:
The modified end of the previous literal is assigned to the literal next to abbrevion. The test of the abbreviations in TestSourcePosition.cs passes only because there is an abbreviation at the end of the string, and thus no literal remains after it. But add some literals after the abbreviation, say, a period '.', and instead of expected literal with the span 103-103 there will be one with the end less then the start (103-98).

### DIFF
--- a/src/Markdig.Tests/TestSourcePosition.cs
+++ b/src/Markdig.Tests/TestSourcePosition.cs
@@ -533,13 +533,17 @@ literal      ( 1, 6)  8-9
     [Test]
     public void TestAbbreviations()
     {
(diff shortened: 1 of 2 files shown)


xoofx:
> But add some literals after the abbreviation, say, a period '.', and instead of expected literal with the span 103-103 there will be one with the end less then the start (103-98).

Could you modify the said test to surface this problem and we have a coverage for this?

snnz:
I've added a paragraph with the abbreviation at the start of the line. It gives "literal (4,4) 111-106" in the current version.

PR #837, opened 2024-12-21
Fix errors in LinkHelper and LinkInlineParser.

### LINKED ISSUE #830: Multiple errors in LinkHelper
Take a look at the examples in the attached file.

### Not recognized as a link:

```
[link](/uri "a
\ \
title")
```

### Recognized as a link reference definition, although it should not be:

```
[foo]: /url (title) x

[foo]
```

### All escaped backslashes in the title (no matter how many) are reduced to one:

```
[link](/uri "\\\\127.0.0.1")
```

snnz:
Here is what I used to fix #830. `LinkHelper.TryParseTitle` is replaced with more structured and clearer one, check it out.

### DIFF
--- a/src/Markdig.Tests/TestPlayParser.cs
+++ b/src/Markdig.Tests/TestPlayParser.cs
@@ -47,6 +47,14 @@ public class TestPlayParser
     }
 
     [Test]
+    public void TestLinkWithMultipleBackslashesInTitle()
+    {
+        var doc = Markdown.Parse(@"[link](/uri '\\\\127.0.0.1')");
+        var link = doc.Descendants<ParagraphBlock>().SelectMany(x => x.Inline.Descendants<LinkInline>()).FirstOrDefault();
+        Assert.AreEqual(@"\\127.0.0.1", link?.Title);
+    }
+
+    [Test]
     public void TestListBug2()
     {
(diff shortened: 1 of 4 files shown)


src/Markdig.Tests/TestPlayParser.cs
```diff
@@ -46,6 +46,14 @@ public void TestLink()
         Assert.AreEqual("/yoyo", link?.Url);
     }
 
+    [Test]
+    public void TestLinkWithMultipleBackslashesInTitle()
+    {
+        var doc = Markdown.Parse(@"[link](/uri '\\\\127.0.0.1')");
+        var link = doc.Descendants<ParagraphBlock>().SelectMany(x => x.Inline.Descendants<LinkInline>()).FirstOrDefault();
```
MihaZupan, line 53:
```suggestion
        var link = doc.Descendants<LinkInline>().FirstOrDefault();
```

snnz:
I've applied the suggestions.

By the way, since backshashes are already handled by `LinkHelper`, and there is fortunately that option, `removeBackSlash`, I didn't touch `HtmlHelper.Unscape` itself. But it is used somewhere else, and doesn't look right: 1) it's this function that eats all the backslashes in a sequence except one, and 2) method `CharHelper.IsEscapableSymbol` that it uses to determine whether to unescape a character or to preserve the backslash, for some reason adds '•' as escapable symbol. Why is that?

PR #838, opened 2024-12-31
Implemented better indent control in TextRendererBase

Melodi17:
- Added method for clearing all indents
- Added logic in PopIndent for when indents are below 1 (TODO mentioned implementing this)

### DIFF
--- a/src/Markdig/Renderers/TextRendererBase.cs
+++ b/src/Markdig/Renderers/TextRendererBase.cs
@@ -1,5 +1,5 @@
 // Copyright (c) Alexandre Mutel. All rights reserved.
-// This file is licensed under the BSD-Clause 2 license. 
+// This file is licensed under the BSD-Clause 2 license.
 // See the license.txt file in the project root for more information.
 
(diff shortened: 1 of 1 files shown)


src/Markdig/Renderers/TextRendererBase.cs
```diff
@@ -174,10 +174,12 @@ public void PushIndent(string[] lineSpecific)
 
     public void PopIndent()
     {
-        // TODO: Check
-        indents.RemoveAt(indents.Count - 1);
+        if (this.indents.Count > 0)
```
xoofx, line 177: This is not better, because it hides the fact that there are more PopIndent than push, which is a bug, so this PR hides bugs. The `TODO: Check` was to say "provide a better error message" instead of failing with list removal.

Melodi17: Ah, my bad sorry. I've added an exception to be thrown when it is empty.

Melodi17: (Sorry for the late response)

PR #851, opened 2025-02-24
Replace encoding polyfill with NET5+ one.

Akarinnnnn:
netstandard2.1 is a special TFM that .NET5+ doesn't mark themselves compitable, even if they mostly are.


MihaZupan (approved):
Thanks, looks like this is the only such ifdef in the repo (a bunch of them could be simplified, but don't change compilation).

Shouldn't matter for perf since the inbox implementation is the same
https://github.com/dotnet/runtime/blob/2a51ee3fd10c2a5af5d44e710ef804813f73bf75/src/libraries/System.Private.CoreLib/src/System/Text/UTF8Encoding.cs#L365-L369

Issue #857, opened 2025-03-15
Add option to AutolinkOptions to allow domains that don't contain a `.`, e.g. `localhost`


JamesNK:
I found https://github.com/xoofx/markdig/issues/827 which has people wanting localhost autolinked. It was closed without helping people support this scenario.

If autolinking http://localhost isn't enabled by default, there should be an option on `AutolinkOptions` to allow it. The alternative requires folks to copy and past all of the autolink code and modify it allow more URLs.

xoofx:
> If autolinking http://localhost isn't enabled by default, there should be an option on `AutolinkOptions` to allow it.

Agreed. Can be added.

JamesNK:
I can make a PR. What kind of option do you want? A bool flag to allow single label domains?

```cs
var options = new AutolinkOptions { AllowSingleLabelDomains = true };
```

PR #859, opened 2025-03-16
Add AutoLinkOptions.AllowDomainWithoutPeriod

JamesNK:
Fixes https://github.com/xoofx/markdig/issues/857

### DIFF
--- a/src/Markdig/Helpers/LinkHelper.cs
+++ b/src/Markdig/Helpers/LinkHelper.cs
@@ -1006,7 +1006,7 @@ public static class LinkHelper
         return c == '\0' || c.IsSpaceOrTab() || c.IsControl() || (isAutoLink && c == '<'); // TODO: specs unclear. space is strict or relaxed? (includes tabs?)
     }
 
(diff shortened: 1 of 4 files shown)


src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -1006,7 +1006,7 @@ private static bool IsEndOfUri(char c, bool isAutoLink)
         return c == '\0' || c.IsSpaceOrTab() || c.IsControl() || (isAutoLink && c == '<'); // TODO: specs unclear. space is strict or relaxed? (includes tabs?)
     }
 
-    public static bool IsValidDomain(string link, int prefixLength)
+    public static bool IsValidDomain(string link, int prefixLength, bool allowDomainWithoutPeriod)
```
MihaZupan, line 1009:
```suggestion
    public static bool IsValidDomain(string link, int prefixLength, bool allowDomainWithoutPeriod = false)
```

To keep it source compatible since these are public helpers

MihaZupan (approved):

xoofx:
Thanks!

PR #864, opened 2025-04-14
A couple perf improvements

MihaZupan:
Improves parsing perf a bit (around 2-4 %).

Logically separated in commits:
- Removing branches from various character checks (removing redundant conditions in calling code, updating helpers)
- A bit faster AutoLink and FencedCodeBlock matching
- Removed per-parsing allocation of delegate captures in AutoIdentifiers
- Using frozen collections in a few minor places


xoofx:
Thanks for the perf improvements!

PR #867, opened 2025-04-15
Update to CommonMark 0.31.2

MihaZupan:
Main changes:
- Definition of "Unicode punctuation character" has changed to also include symbol categories.
  - We were using the same helper when parsing autolinks (`<http://...>`) and GFM autolinks (`http://...`)
  - The one for AutoLinks seems to have been wrong, updated it to match the spec now
  - GitHub appears to allow symbols in its autolinks (e.g. emoji in domain name), so I tried to match that by keeping the old behavior there
- More forms of inline raw HTML comments are allowed: `<!-->`, `<!--->`
- `search` is apparently an HTML tag now


xoofx:
Amazing, thanks for the update! 🚀

Issue #873, opened 2025-05-25
Crashing on iOS and Android after usage over 1000 times


taublast:
EDIT: in context of the final repro:

I have a net9.0-android net9.0-ios MAUI app that uses Markding for chat list bubbles labels. This works fine until cells call `Parse` method over 1000 times when scrolling recycled cells.

Removing those lines from Markdig fork made crashes to disappear while being able to call `Parse` when Markdig project is referenced in solution and compiled for Debug:

```csharp
#if NET5_0_OR_GREATER
[module: System.Runtime.CompilerServices.SkipLocalsInit]
#endif
```

This fixed crashing when Markdig fork is refereced as project. When it's compiled for Release crashes come back again.

Those native crashes obviously can't be catched in debugger becaus they happen inside .NET optimized binaries.

Crashes:

Android:

```
05-24 21:08:12.399 19548 19548 D ViewRootImpl@332085e[MainActivity]: ViewPostIme pointer 1
05-24 21:08:12.419 19628 19628 E crash_dump64: unknown process state: t
05-24 21:08:12.452 19628 19628 I crash_dump64: obtaining output fd from tombstoned, type: kDebuggerdTombstone
05-24 21:08:12.453  1142  1142 I /system/bin/tombstoned: received crash request for pid 19584
05-24 21:08:12.454 19628 19628 I crash_dump64: performing dump of process 19548 (target tid = 19584)
05-24 21:08:12.463 19628 19628 F DEBUG   : *** *** *** *** *** *** *** *** *** *** *** *** *** *** *** ***
05-24 21:08:12.463 19628 19628 F DEBUG   : Build fingerprint: 'samsung/beyond2qltesq/beyond2q:10/QP1A.190711.020/G975USQS4DTG1:user/release-keys'
05-24 21:08:12.463 19628 19628 F DEBUG   : Revision: '17'
05-24 21:08:12.463 19628 19628 F DEBUG   : ABI: 'arm64'
05-24 21:08:12.463 19628 19628 F DEBUG   : Timestamp: 2025-05-24 21:08:12+0300
05-24 21:08:12.463 19628 19628 F DEBUG   : pid: 19548, tid: 19584, name: GLThread 6  >>> com.appomobi.app.dev <<<
05-24 21:08:12.463 19628 19628 F DEBUG   : uid: 10303
05-24 21:08:12.463 19628 19628 F DEBUG   : signal 11 (SIGSEGV), code 1 (SEGV_MAPERR), fault addr 0x8
[...]

PR #874, opened 2025-05-25
Remove module-level [SkipLocalsInit] to prevent native crashes on AOT platforms (iOS/Android)

taublast:
This PR removes the following directive:

```csharp
#if NET5_0_OR_GREATER
[module: System.Runtime.CompilerServices.SkipLocalsInit]
#endif
```

to resolve native crashes on mobile platforms (specifically catched on iOS and Android when using .NET 9 in MAUI app).

🧯 The Problem

In high-throughput UI scenarios (e.g. extremely fast list scrolling in MAUI), calls to:

```csharp
Markdown.Parse(string, MarkdownPipeline)
```

began randomly triggering crashes. After isolating the issue, removing [SkipLocalsInit] fully resolved the problem.
Removing it ensures cross-platform safety and improves compatibility mobile/AOT targets.

Issue describing the case: https://github.com/xoofx/markdig/issues/873

"This attribute is unsafe, because it may reveal uninitialized memory to the application in certain instances (for example, reading from uninitialized stack-allocated memory). If applied to a method directly, the attribute applies to that method and all its nested functions, including lambdas and local functions. If applied to a type or module, it applies to all methods nested inside. This attribute is intentionally not permitted on assemblies. To apply the attribute to multiple type declarations, use it at the module level instead."
— [docs.microsoft.com](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.skiplocalsinitattribute)

🚀 Impact
Fixes crashes in mobile apps using Markdig.  Improves confidence for developers using Markdig in MAUI, Unity, Blazor Hybrid, and other non-Windows environments.  

Please let me know if you'd prefer the attribute be conditionally compiled or applied to only select methods, would happy to adjust.

### DIFF
--- a/src/Markdig/SkipLocalsInit.cs
+++ b/src/Markdig/SkipLocalsInit.cs
@@ -1,7 +1,0 @@
-// Copyright (c) Alexandre Mutel. All rights reserved.
-// This file is licensed under the BSD-Clause 2 license.
-// See the license.txt file in the project root for more information.
-
-#if NET5_0_OR_GREATER
-[module: System.Runtime.CompilerServices.SkipLocalsInit]
-#endif


taublast:
Yes, you are totally right. Removing this makes it works as referenced project on Debug. As soon as i compiled the custom nuget on Release the crash came back again. Tried to change `StackallocThreshold` no luck.. 
Still investigating, something related to GC hitting over spans in the middle of the Markdig processing.
Will pull out a repro in the upcoming days.
Will close the PR and would share more info inside the replated issue.

PR #876, opened 2025-05-31
Fix #872 by reserve null title string.

### LINKED ISSUE #872: Get `LinkInline.Title` of `""` rather than `null` for `[link](/uri)`
We get `LinkInline.Title` of `""` with [CommonMark Example 443](https://spec.commonmark.org/0.22/#example-443).

> `[link](/uri)`

As `LinkInline.Title` is typed as `string?` rather than `string`, it is weird.

https://github.com/xoofx/markdig/blob/3535701d70066d055d06adaeaa1531b10601c6ac/src/Markdig/Renderers/Html/Inlines/LinkInlineRenderer.cs#L72-L77

`HtmlRenderer` render correct result for `[link](/uri)` with current parse behavior, but it will generate incorrect result for `[link](/uri "")` instead.

Akarinnnnn:
#872 reported an inconsistency between nullable annotation and possible value. In case of no title exists in a link syntax, consumers except `Title` property returns a null string. But current behavior is `Title` returning an empty string, This PR fixes the inconsistency by make `Title` property null if link title is not exist.

### DIFF
--- a/src/Markdig/Helpers/HtmlHelper.cs
+++ b/src/Markdig/Helpers/HtmlHelper.cs
@@ -470,6 +470,24 @@ public static class HtmlHelper
     /// </summary>
     /// <param name="text">The string data that will be changed by unescaping any punctuation or symbol characters.</param>
     /// <param name="removeBackSlash">if set to <c>true</c> [remove back slash].</param>
+    /// <returns>Unescaped text, or null if <paramref name="text"/> is null.<returns>
+    public static string? UnescapeNullable(string? text, bool removeBackSlash = true)
+    {
+        if (text == null)
+        {
+            return null;
+        }
+        else
+        {
+            return Unescape(text, removeBackSlash);
+        }
+    }
+
+    /// <summary>
+    /// Destructively unescape a string: remove backslashes before punctuation or symbol characters.
+    /// </summary>
+    /// <param name="text">The string data that will be changed by unescaping any punctuation or symbol characters.</param>
+    /// <param name="removeBackSlash">if set to <c>true</c> [remove back slash].</param>
     /// <returns></returns>
     public static string Unescape(string? text, bool removeBackSlash = true)
     {
--- a/src/Markdig/Helpers/HtmlHelper.cs
+++ b/src/Markdig/Helpers/HtmlHelper.cs
@@ -470,24 +470,6 @@ public static class HtmlHelper
     /// </summary>
     /// <param name="text">The string data that will be changed by unescaping any punctuation or symbol characters.</param>
     /// <param name="removeBackSlash">if set to <c>true</c> [remove back slash].</param>
-    /// <returns>Unescaped text, or null if <paramref name="text"/> is null.<returns>
-    public static string? UnescapeNullable(string? text, bool removeBackSlash = true)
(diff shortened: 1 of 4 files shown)


src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -411,7 +411,7 @@ public static bool TryParseInlineLink(ref StringSlice text, out string? link, ou
         {
             // Skip ')'
             text.SkipChar();
-            title ??= string.Empty;
+            // not to normalize nulls
```
MihaZupan, line 414: This comment makes sense if you're looking at this PR, but not on its own. I'd just remove it

src/Markdig/Helpers/HtmlHelper.cs
```diff
@@ -465,6 +465,24 @@ private static bool TryParseHtmlTagProcessingInstruction(ref StringSlice text, r
         }
     }
 
+    /// <summary>
+    /// Destructively unescape a string: remove backslashes before punctuation or symbol characters.
+    /// </summary>
+    /// <param name="text">The string data that will be changed by unescaping any punctuation or symbol characters.</param>
+    /// <param name="removeBackSlash">if set to <c>true</c> [remove back slash].</param>
+    /// <returns>Unescaped text, or null if <paramref name="text"/> is null.<returns>
+    public static string? UnescapeNullable(string? text, bool removeBackSlash = true)
```
MihaZupan, line 474: I don't think this helper is valuable, can we do a simple null check at the one call site instead?

src/Markdig/Helpers/LinkHelper.cs
Akarinnnnn: Changed, explained why doing so in detail.

Akarinnnnn:
Sounds good, I misunderstood the meaning of callsite.



---Original---
From: "Miha ***@***.***&gt;
Date: Sat, May 31, 2025 23:23 PM
To: ***@***.***&gt;;
Cc: ***@***.******@***.***&gt;;
Subject: Re: [xoofx/markdig] Fix #872 by reserve null title string. (PR #876)




 
@MihaZupan commented on this pull request.
 
 
In src/Markdig/Helpers/HtmlHelper.cs:
 &gt; @@ -465,6 +465,24 @@ private static bool TryParseHtmlTagProcessingInstruction(ref StringSlice text, r          }      }   +    /// <summary&gt; +    /// Destructively unescape a string: remove backslashes before punctuation or symbol characters. +    /// </summary&gt; +    /// <param name="text"&gt;The string data that will be changed by unescaping any punctuation or symbol characters.</param&gt; +    /// <param name="removeBackSlash"&gt;if set to <c&gt;true</c&gt; [remove back slash].</param&gt; +    /// <returns&gt;Unescaped text, or null if <paramref name="text"/&gt; is null.<returns&gt; +    public static string? UnescapeNullable(string? text, bool removeBackSlash = true)  
I'm not suggesting to change the behavior of Unescape, but to change the one call site
 -Title = HtmlHelper.UnescapeNullable(title, removeBackSlash: false), +Title = title is null ? null : HtmlHelper.Unescape(title, removeBackSlash: false),

PR #883, opened 2025-07-14
Add source position tracking for grid tables

### LINKED ISSUE #881: Incorrect Line/Column positions inside GridTable
I have the following markdown:

```
XXX

+---+-------------------------------------------------------------------+
| A | B                                                                 |
+=======================================================================+
| C | [LinkA](http://example.com) |
+---+-------------------------------------------------------------------+
| D | E                                                                 |
+---+-------------------------------------------------------------------+

a | b
-- | -
0 | [LinkB](http://example.com)


YYY
```

The line and column reported for the link inside the PipeTable (LinkB) are correct: Line 12, Column 4.
However, the link inside the GridTable (LinkA) is incorrectly reported at Line 5, Column 1.

I tried to fix this issue, but since I had to spend some time learning how the code works, my fix is probably not clean or stable enough for a pull request.
You can find it here, in my forked repository:
https://github.com/messani/markdig

Can you please fix it?

messani:
Fixes #881

Adds source position tracking for grid tables and corresponding tests.

### DIFF
--- a/src/Markdig/Parsers/BlockProcessor.cs
+++ b/src/Markdig/Parsers/BlockProcessor.cs
@@ -485,15 +485,21 @@ public class BlockProcessor
     /// Processes a new line.
     /// </summary>
     /// <param name="newLine">The new line.</param>
-    public void ProcessLine(StringSlice newLine)
+    /// <param name="column">The offset.</param>
+    public void ProcessLine(StringSlice newLine, int column = 0)
     {
         CurrentLineStartPosition = newLine.Start;
 
-        Document.LineStartIndexes?.Add(CurrentLineStartPosition);
+        if (column == 0)
+        {
+            Document.LineStartIndexes?.Add(CurrentLineStartPosition);
+        }
 
         ContinueProcessingLine = true;
 
         ResetLine(newLine);
+        Column = column;
+        originalLineStart -= column;
 
         TryContinueBlocks();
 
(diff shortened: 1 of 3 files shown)


src/Markdig/Parsers/BlockProcessor.cs
```diff
@@ -485,25 +485,53 @@ public void Discard(Block block)
...
 
-        Document.LineStartIndexes?.Add(CurrentLineStartPosition);
+        if (column == 0)
+        {
+            Document.LineStartIndexes?.Add(CurrentLineStartPosition);
+        }
+
+        ContinueProcessingLine = true;
+
+        ResetLine(newLine, 0);
```
xoofx, line 496:
```suggestion
        ResetLine(newLine, column);
```

Shouldn't it pass column?

messani:
Sorry about that — the first fix for this issue was a bit of a hack, and part of that code was left behind. It shouldn't have affected functionality, but it wasn't clean. I've corrected it now, and it should be fine. Thanks for spotting it!

messani:
Since I'm not the original author, my contribution might not be entirely ideal.
The original ProcessLine method works the same way as before.
I've introduced a new method, ProcessLinePart, which doesn't add an entry to the LineStartIndexes variable and doesn't increment LineIndex.
Unfortunately, I'm not entirely sure how the BlockProcessor is designed internally, so I hope this change doesn't cause any issues.

xoofx:
Thanks for this. As I have been quite away from the details on this project, I will have to double check things, but it looks ok. 🙂

Issue #818, opened 2024-09-28
Tables are not rendered properly unless there is a new line before them


dusanrad:
Consider a following piece of code:
``` c#
using Markdig;

var input = @"
Here is the table:
| Key   | Value    |
| ----  | -------- |
| Color | Blue     |
| Size  | Medium   |
";

var pipeline = new MarkdownPipelineBuilder().UsePipeTables().Build();
var html = Markdown.ToHtml(input, pipeline);
Console.WriteLine(html);
```

It prints single paragraph without converting the table to html:
``` html
<p>Here is the table:
| Key   | Value    |
| ----  | -------- |
| Color | Blue     |
| Size  | Medium   |</p>
```

If we just add new line after the `Here is the table:` part:

``` c#
using Markdig;

var input = @"
Here is the table:

| Key   | Value    |
| ----  | -------- |
| Color | Blue     |
| Size  | Medium   |
";

var pipeline = new MarkdownPipelineBuilder().UsePipeTables().Build();
var html = Markdown.ToHtml(input, pipeline);
Console.WriteLine(html);
```

then everything gets converted properly:
``` html
<p>Here is the table:</p>
<table>
<thead>
<tr>
<th>Key</th>
<th>Value</th>
</tr>
</thead>
<tbody>
<tr>
<td>Color</td>
<td>Blue</td>
</tr>
<tr>
<td>Size</td>
<td>Medium</td>
</tr>
</tbody>
</table>
```

This is different from how other Markdown libraries are rendering it. It is also different from how Github renders it. If I just paste Markdown code from the first example in the text of this issue we properly rendered table:

Here is the table:
| Key   | Value    |
| ----  | -------- |
| Color | Blue     |
| Size  | Medium   |

Let me know if more information is needed.

xoofx:
When markdig was implemented, it was based on pandoc behavior. The parsing of pipe tables is a complicated beast, and as this library is mainly in maintenance mode, I won't take the time to fix it, but if someone is brave enough, a PR is welcome.

snnz:
There is explicit rule in the [specs](https://github.com/xoofx/markdig/blob/master/src/Markdig.Tests/Specs/PipeTableSpecs.md):
> But if a table doesn't start with a column delimiter, it is not interpreted as a table, even if following lines have a column delimiter

But I can't find where it could possibly come from in the pandoc specs.

GFM is different, and not only in this aspect:
[...]

PR #885, opened 2025-08-09
Add support for a table without an extra new line before it

dannyp32:
Fixes the following issue where a table will not be rendered if there isn't an extra line before the table. 
https://github.com/xoofx/markdig/issues/818

For example this will not render correctly

```
Here is the table:
| Key   | Value    |
| ----  | -------- |
| Color | Blue     |
| Size  | Medium   |
```
But this will because there's an extra new line before the table starts

```
Here is the table:

| Key   | Value    |
| ----  | -------- |
| Color | Blue     |
| Size  | Medium   |
```

Not handled in this change - whitespace between the new line and the pipe

### DIFF
--- a/src/Markdig.Tests/TestPipeTable.cs
+++ b/src/Markdig.Tests/TestPipeTable.cs
@@ -10,6 +10,8 @@ public sealed class TestPipeTable
     [TestCase("| S | T |\r\n|---|---|\t\r\n| G | H |")]
     [TestCase("| S | T |\r\n|---|---|\f\r\n| G | H |")]
     [TestCase("| S | \r\n|---|\r\n| G |\r\n\r\n| D | D |\r\n| ---| ---| \r\n| V | V |", 2)]
+    [TestCase("a\r| S | T |\r|---|---|")]
+    [TestCase("a\n| S | T |\r|---|---|")]
     public void TestTableBug(string markdown, int tableCount = 1)
     {
         MarkdownDocument document =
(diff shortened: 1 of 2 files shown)


dannyp32:
@xoofx hoping you can take a look. Open to any suggestions

xoofx:
Thanks for this. I will have to double check some stuffs but it could be enough.

RickStrahl:
Checking in 0.42.0 - looks like this works for Grid Tables but not for Pipe Tables.

image

MihaZupan:
Possibly an issue with how we're dealing with new lines in this change.
The PR added tests for `a\r|...` and `a\n|...` - if I change that to `\r\n` instead it fails.

MihaZupan:
Looks like this change is also removing the content before the table
```c#
string md =
    """
    Some text
    | A |
    |---|
    | B |
    """.ReplaceLineEndings("\n");
    
    .UsePipeTables()
    .Build();

Console.WriteLine(Markdown.ToHtml(md, pipeline));
```
(note the missing "some text")
```md
<table>
<thead>
<tr>
<th>A</th>
</tr>
</thead>
<tbody>
<tr>
<td>B</td>
</tr>
</tbody>
</table>
```

And it doesn't match the table if the text before it is more than one line (not sure if this was intentional).

RickStrahl:
FWIW - I normalize my linefeeds to LF before sending to Markdig, so that's not the issue I'm seeing...

MihaZupan:
I suspect that your case hit this
> And it doesn't match the table if the text before it is more than one line (not sure if this was intentional).

Should be fixed with #905 to allow multiple lines of text to precede the table.

Issue #845, opened 2025-01-03
A list with empty lines extends into the next paragraph.


snnz:
Noticed this while looking at the test samples in #839.

CommonMark spec states that a list item can begin with at most one blank line. [Example 280](https://spec.commonmark.org/0.31.2/#example-280) demonstrates this.
```csharp
var result = Markdown.ToHtml("-\n\n  foo");
```

The result is:

```
<ul>
<li></li>
</ul>
<p>foo</p>
```

If one more empty item is added

```csharp
var result = Markdown.ToHtml("-\n-\n\n  foo");
```

Everything is as expected again:

```
<ul>
<li></li>
<li></li>
</ul>
<p>foo</p>
```

But if a blank line is inserted between the items:

```csharp
var result = Markdown.ToHtml("-\n\n-\n\n  foo");
```

The result is suddenly

```
<ul>
<li></li>
<li><p>foo</p></li>
</ul>
```

Isn't this strange?

xoofx:
Repro [here](https://babelmark.github.io/?text=-%0A%0A-%0A%0A++foo%0A)

> Isn't this strange?

Yes, it's probably a bug.

MihaZupan:
This one presumably got fixed by #888 with the 0.42.0 release.

PR #888, opened 2025-09-06
Fixes issue #845

prozolic:
This PR fixes the bug reported in issue #845.
With this fix, in the following case:

```csharp
var result = Markdown.ToHtml("-\n\n-\n\n  foo");
```

This result will be produced:
```
<ul>
<li></li>
<li></li>
</ul>
<p>foo</p>
```


xoofx:
Thanks for the fix! Could you add a unit test?

prozolic:
Ok, I added the test.

Issue #504, opened 2021-01-21
Depth limit exceeded on non-malicious markdown content


yufeih:
[This file](https://github.com/microsoftgraph/microsoft-graph-docs/blob/master/api-reference/beta/api/intune-deviceconfig-macosdevicefeaturesconfiguration-create.md) raises depth limit exceeded exception in version 0.23.0 due to #500.

Some additional info:
- That file is an auto generated API reference page.
- I cannot give a reasonable depth limit that don't break all content in docs.microsoft.com, but a depth limit configuration is a good start to mitigate the problem.
- I did some perf profiling of markdig on a bunch of real markdown files used by docs.microsoft.com and the hottest function inside markdig is [`FindLastContainer`](https://github.com/xoofx/markdig/pull/500/files#diff-a3f6a991f6aa7df0c1a521fb37af183e8784d46fc93beb770e00642501983916R307). I don't have enough time to dig deeper, not sure if there are faster ways to find the last container (like tracking its location instead of search the tree each time?)

MihaZupan:
Just looking at the file, we shouldn't be approaching the current 10k limit 🤔 

I will look into what is happening here this week.

MihaZupan:
Something weird is happening with table parsing here.

Whenever we have single backticks '\`' in different lines, Markdig will give up on the table and parse it into a very deep model - **one nested level for each pipe character!**

Example demonstrating this: https://gist.github.com/MihaZupan/98621af5788f570f0dd0b809f1f3489b
Note the "BadTable" being rather small, but with 2 extra backticks - this will fail to parse as a table in Markdig.
The "GoodTable", approaching but under 10k cells, will parse as a table just fine.

You can see that this is happening for the sample file you mentioned. If you browse to the rendered docs page, you will see that it's completely broken:
https://docs.microsoft.com/en-us/graph/api/intune-deviceconfig-macosdevicefeaturesconfiguration-create?view=graph-rest-beta#request-body


In the sample file, the problem is caused by the standalone \` in this line of the table (left of "console"):

|consoleAccessDisabled|Boolean|Whether the Other user will disregard use of the `>console> special user name.|
|-|-|-|

So the issue for that specific file can be solved by just removing the extra backtick, but we'll have to investigate what's going on overall.

yufeih:
I could forward to the relevant folks to see if we could escape the problematic ` in the API generation process. GitHub seems to be rendering that table just fine, I'm not sure if this is another difference between GFM and CommonMark.

[...]

PR #891, opened 2025-09-18
Fix for table depth error when cell contains backticks

phillip-haydon:
This addresses the issue in #504 using the example from https://github.com/xoofx/markdig/issues/504#issuecomment-765049325

https://gist.github.com/MihaZupan/98621af5788f570f0dd0b809f1f3489b

Not sure what the process is on contributing as there's no open PRs so I don't think this has been fixed before :)

### DIFF
--- a/src/Markdig/Extensions/Tables/PipeTableParser.cs
+++ b/src/Markdig/Extensions/Tables/PipeTableParser.cs
@@ -196,6 +196,16 @@ public class PipeTableParser : InlineParser, IPostInlineProcessor
         // Continue
         if (tableState is null || container is null || tableState.IsInvalidTable || !tableState.LineHasPipe ) //|| tableState.LineIndex != state.LocalLineIndex)
         {
+            if (tableState is not null)
+            {
+                foreach (var inline in tableState.ColumnAndLineDelimiters)
+                {
+                    if (inline is PipeTableDelimiterInline pipeDelimiter)
+                    {
+                        pipeDelimiter.ReplaceByLiteral();
+                    }
+                }
+            }
             return true;
         }
 
--- a/src/Markdig/Parsers/Inlines/CodeInlineParser.cs
+++ b/src/Markdig/Parsers/Inlines/CodeInlineParser.cs
@@ -35,6 +35,7 @@ public class CodeInlineParser : InlineParser
         Debug.Assert(match is not ('\r' or '\n'));
 
         // Match the opened sticks
+        int openingStart = slice.Start;
         int openSticks = slice.CountAndSkipChar(match);
 
         // A backtick string is a string of one or more backtick characters (`) that is neither preceded nor followed by a backtick.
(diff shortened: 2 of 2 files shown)


xoofx:
Thanks for the fix!

It could be an ok fix. I will try to have a look locally. My only concern is that it is coupling the code inline parser with a potential grid table `|` character, but yeah, not sure there are much cleaner way.

xoofx:
cc: @MihaZupan if you have any comments on this

phillip-haydon:
Thanks for accepting the PR ❤️ do you know when a new nuget would be available? I'm just running my app with a local nuget dir at the moment but would like ti get it from nuget.org when available.

src/Markdig/Extensions/Tables/PipeTableParser.cs
```diff
@@ -196,6 +196,16 @@ public bool PostProcess(InlineProcessor state, Inline? root, Inline? lastChild,
...
+            if (tableState is not null)
+            {
+                foreach (var inline in tableState.ColumnAndLineDelimiters)
+                {
+                    if (inline is PipeTableDelimiterInline pipeDelimiter)
+                    {
+                        pipeDelimiter.ReplaceByLiteral();
+                    }
+                }
+            }
```
MihaZupan, line 208: This is merely cleanup so that the syntax tree doesn't have random pipe delimiters even if we didn't have a table in the end, right?

None of the tests seem to care since the literal renders the same either way.

MihaZupan (commented):
Table parsing is the one area of Markdig I'm not that familiar with unfortunately.

I fear this change is going to break otherwise valid inputs, e.g.
```
`
|| hidden text ||
`
```
is currently a perfectly valid `CodeInline` ([and all parsers agree on that](https://babelmark.github.io/?text=%60%0A%7C%7C+hidden+text+%7C%7C%0A%60)), but this change turns it into a literal instead.

Since this is clearly a targeted fix for table weirdness, I wonder if we should embrace that and add a `&& processor.Inline?.ContainsParentOfType<PipeTableDelimiterInline>() == true` check before breaking the `CodeInline`?

phillip-haydon:
> Table parsing is the one area of Markdig I'm not that familiar with unfortunately.
> 
> I fear this change is going to break otherwise valid inputs, e.g.
> 
> ```
> `
> || hidden text ||
> `
> ```
> 
> is currently a perfectly valid `CodeInline` ([and all parsers agree on that](https://babelmark.github.io/?text=%60%0A%7C%7C+hidden+text+%7C%7C%0A%60)), but this change turns it into a literal instead.
> 
> Since this is clearly a targeted fix for table weirdness, I wonder if we should embrace that and add a `&& processor.Inline?.ContainsParentOfType<PipeTableDelimiterInline>() == true` check before breaking the `CodeInline`?

I tested:

```
        const string markdown = "`|| hidden text ||`";
```
and
```
        const string markdown = """
                                `
                                || hidden text ||
                                `
                                """;
```

Both produced:

```
<p><code>|| hidden text ||</code></p>
```

src/Markdig/Extensions/Tables/PipeTableParser.cs
phillip-haydon: Yeah it's just cleanup, it swaps the left over | nodes back to plain text so the tree looks tidy. I did run the benchmark against it and there was no impact on perf so kept it in.

phillip-haydon:
Would you like me to add a follow up PR with unit tests for the code snippets?

MihaZupan:
The second case (with new lines) isn't a code inline anymore after your change.
```c#
[Test]
public void Test()
{
    TestParser.TestSpec(
        """
        `
        || hidden text ||
        `
        """,
        "<p><code> || hidden text || </code></p>");
}
```
fails with
```
       ```````````````````Source
       `
       ||·hidden·text·||
       `
       ```````````````````Result
       <p>`
       ||·hidden·text·||
       `</p>
       ```````````````````Expected
       <p><code>·||·hidden·text·||·</code></p>
       ```````````````````
```

More unit tests could be useful (e.g. to also check cases with leading spaces), but I think we should tweak behavior here.

phillip-haydon:
> The second case (with new lines) isn't a code inline anymore after your change.
> 
> ```cs
> [Test]
> public void Test()
> {
>     TestParser.TestSpec(
>         """
>         `
>         || hidden text ||
>         `
>         """,
>         "<p><code> || hidden text || </code></p>");
> }
> ```
> 
> fails with
> 
> ```
>        ```````````````````Source
>        `
>        ||·hidden·text·||
>        `
>        ```````````````````Result
>        <p>`
>        ||·hidden·text·||
>        `</p>
>        ```````````````````Expected
>        <p><code>·||·hidden·text·||·</code></p>
>        ```````````````````
> ```
> 
> More unit tests could be useful (e.g. to also check cases with leading spaces), but I think we should tweak behavior here.

OH, ok leave that to me, ill get a unit test in and fix tonight.

PR #893, opened 2025-09-22
Fix issue where an inline code block that spans multiple lines doesn't parse correctly

phillip-haydon:
Based on feedback from PR #891 - this PR fixes inline code block that spans multiple lines.

```
`
|| hidden text ||
`
```

### DIFF
--- a/src/Markdig.Tests/TestPipeTable.cs
+++ b/src/Markdig.Tests/TestPipeTable.cs
@@ -1,5 +1,7 @@
+using Markdig;
 using Markdig.Extensions.Tables;
 using Markdig.Syntax;
+using Markdig.Syntax.Inlines;
 
 namespace Markdig.Tests;
 
@@ -100,5 +102,73 @@ public sealed class TestPipeTable
         Assert.DoesNotThrow(() => html = Markdown.ToHtml(markdown, pipeline));
         Assert.That(html, Does.Contain("<table"));
         Assert.That(html, Does.Contain("<td>`C</td>"));
+    }
+
+    [Test]
+    public void CodeInlineWithPipeDelimitersRemainsCodeInline()
+    {
+        const string markdown = "`|| hidden text ||`";
+
+        var pipeline = new MarkdownPipelineBuilder()
+            .UseAdvancedExtensions()
+            .Build();
+
+        var document = Markdown.Parse(markdown, pipeline);
+
+        var codeInline = document.Descendants().OfType<CodeInline>().SingleOrDefault();
+        Assert.IsNotNull(codeInline);
+        Assert.That(codeInline!.Content, Is.EqualTo("|| hidden text ||"));
+    }
+
+    [Test]
+    public void SingleLineCodeInlineWithPipeDelimitersRendersAsCode()
+    {
+        const string markdown = "`|| hidden text ||`";
+
+        var pipeline = new MarkdownPipelineBuilder()
+            .UseAdvancedExtensions()
+            .Build();
+
+        var html = Markdown.ToHtml(markdown, pipeline);
+
+        Assert.That(html, Is.EqualTo("<p><code>|| hidden text ||</code></p>\n"));
+    }
+
+    [Test]
+    public void MultiLineCodeInlineWithPipeDelimitersRendersAsCode()
+    {
+        const string markdown = """
+`
+|| hidden text ||
+`
+""";
+
+        var pipeline = new MarkdownPipelineBuilder()
+            .UseAdvancedExtensions()
+            .Build();
+
+        var html = Markdown.ToHtml(markdown, pipeline);
+
+        Assert.That(html, Is.EqualTo("<p><code>|| hidden text ||</code></p>\n"));
+    }
+
+    [Test]
+    public void TableCellWithCodeInlineRendersCorrectly()
+    {
+        const string markdown = """
+| Count | A | B | C | D | E |
+|-------|---|---|---|---|---|
+|     0 | B | C | D | E | F |
+|     1 | B | `Code block` | D | E | F |
+|     2 | B | C | D | E | F |
+""";
+
+        var pipeline = new MarkdownPipelineBuilder()
+            .UseAdvancedExtensions()
+            .Build();
+
+        var html = Markdown.ToHtml(markdown, pipeline);
+
+        Assert.That(html, Does.Contain("<td><code>Code block</code></td>"));
     }
 }
--- a/src/Markdig/Extensions/Tables/PipeTableParser.cs
+++ b/src/Markdig/Extensions/Tables/PipeTableParser.cs
@@ -47,6 +47,16 @@ public class PipeTableParser : InlineParser, IPostInlineProcessor
             return false;
         }
 
+        // Do not treat pipe characters that occur within inline code block as table delimiters
+        // This handles the scenario where we start an inline code block but do a new line with a pipe
+        // `
+        // || Code ||
+        // `
+        if (processor.Inline != null && processor.Inline.ContainsParentOfType<CodeInline>())
+        {
+            return false;
+        }
+
         var c = slice.CurrentChar;
         var isNewLineFollowedByPipe = (c == '\n' || c == '\r') && slice.PeekChar() == '|';
 
--- a/src/Markdig/Helpers/HtmlHelper.cs
+++ b/src/Markdig/Helpers/HtmlHelper.cs
@@ -1,5 +1,5 @@
 // Copyright (c) Alexandre Mutel. All rights reserved.
-// This file is licensed under the BSD-Clause 2 license. 
+// This file is licensed under the BSD-Clause 2 license.
 // See the license.txt file in the project root for more information.
 
 using System.Diagnostics;
@@ -430,7 +430,7 @@ public static class HtmlHelper
 
         const string EndOfComment = "-->";
 
-        int endOfComment = slice.IndexOf(EndOfComment, StringComparison.Ordinal);
+        int endOfComment = slice.IndexOf(EndOfComment.AsSpan(), StringComparison.Ordinal);
         if (endOfComment < 0)
         {
             return false;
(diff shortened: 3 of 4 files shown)


src/Markdig/Helpers/HtmlHelper.cs
```diff
@@ -430,7 +430,7 @@ private static bool TryParseHtmlTagHtmlComment(ref StringSlice text, ref ValueSt
 
         const string EndOfComment = "-->";
 
-        int endOfComment = slice.IndexOf(EndOfComment, StringComparison.Ordinal);
+        int endOfComment = slice.IndexOf(EndOfComment.AsSpan(), StringComparison.Ordinal);
```
phillip-haydon, line 433: I included this change because for some reason, on macOS. `EndOfComment` wont compile. It gets:

> 0>HtmlHelper.cs(433,42): Error CS1503 : Argument 2: cannot convert from 'string' to 'System.ReadOnlySpan<char>'

So I've been adding `AsSpan()` but it may make sense to just get this in too? Can revert if you want.

phillip-haydon:
@MihaZupan if you don't mind taking a look at this PR please.

src/Markdig/Extensions/Tables/PipeTableParser.cs
```diff
@@ -47,6 +47,16 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             return false;
         }
 
+        // Do not treat pipe characters that occur within inline code block as table delimiters
+        // This handles the scenario where we start an inline code block but do a new line with a pipe
+        // `
+        // || Code ||
+        // `
+        if (processor.Inline != null && processor.Inline.ContainsParentOfType<CodeInline>())
```
MihaZupan: When I was testing your previous PR locally with this example I didn't need checks here (only in `CodeInlineParser`).
Is there a different test case this is needed for?

Edit: Not seeing any test failures if I delete this

src/Markdig.Tests/TestPipeTable.cs
```diff
@@ -101,4 +103,72 @@ public void TableWithUnbalancedCodeSpanParsesWithoutDepthLimitError()
...
+    {
+        const string markdown = "`|| hidden text ||`";
+
+        var pipeline = new MarkdownPipelineBuilder()
+            .UseAdvancedExtensions()
+            .Build();
+
+        var html = Markdown.ToHtml(markdown, pipeline);
+
+        Assert.That(html, Is.EqualTo("<p><code>|| hidden text ||</code></p>\n"));
```
MihaZupan, line 140:
```suggestion
        Assert.That(docunent.ToHtml(), Is.EqualTo("<p><code>|| hidden text ||</code></p>\n"));
```

src/Markdig.Tests/TestPipeTable.cs
```diff
@@ -101,4 +103,72 @@ public void TableWithUnbalancedCodeSpanParsesWithoutDepthLimitError()
...
+    }
+
+    [Test]
+    public void MultiLineCodeInlineWithPipeDelimitersRendersAsCode()
+    {
+        const string markdown = """
+`
+|| hidden text ||
+`
+""";
```
MihaZupan:
```suggestion
        const string markdown =
            """
            `
            || hidden text ||
            `
            """;
```

src/Markdig.Tests/TestPipeTable.cs
```diff
@@ -101,4 +103,72 @@ public void TableWithUnbalancedCodeSpanParsesWithoutDepthLimitError()
...
+    [Test]
+    public void TableCellWithCodeInlineRendersCorrectly()
+    {
+        const string markdown = """
+| Count | A | B | C | D | E |
+|-------|---|---|---|---|---|
+|     0 | B | C | D | E | F |
+|     1 | B | `Code block` | D | E | F |
+|     2 | B | C | D | E | F |
+""";
```
MihaZupan:
```suggestion
        const string markdown =
            """
            | Count | A | B | C | D | E |
            |-------|---|---|---|---|---|
            |     0 | B | C | D | E | F |
            |     1 | B | `Code block` | D | E | F |
            |     2 | B | C | D | E | F |
            """;
```

src/Markdig.Tests/TestPipeTable.cs
```diff
@@ -101,4 +103,72 @@ public void TableWithUnbalancedCodeSpanParsesWithoutDepthLimitError()
...
+        Assert.That(html, Is.EqualTo("<p><code>|| hidden text ||</code></p>\n"));
+    }
+
+    [Test]
+    public void TableCellWithCodeInlineRendersCorrectly()
+    {
+        const string markdown = """
+| Count | A | B | C | D | E |
+|-------|---|---|---|---|---|
+|     0 | B | C | D | E | F |
```
MihaZupan: Please also add a test case for the case with an indented pipe character in the table.
Something that exercises this added code path:
https://github.com/xoofx/markdig/blob/0e9e80e1cd9c544d77eae59d62efc333bfc0cc3e/src/Markdig/Parsers/Inlines/CodeInlineParser.cs#L88-L91

src/Markdig/Helpers/HtmlHelper.cs
MihaZupan: Was it failing for the Framework/standard builds?

phillip-haydon:
> Was it failing for the Framework/standard builds?

I don't know about framework as I'm on macOS, but it fails for net standard builds.

src/Markdig/Helpers/HtmlHelper.cs
phillip-haydon: I don't know about framework as I'm on macOS, but it fails for net standard builds.

src/Markdig.Tests/TestPipeTable.cs
phillip-haydon: 🤔 why did i add the test twice...

src/Markdig.Tests/TestPipeTable.cs
phillip-haydon: Added test for this.

phillip-haydon: I also removed the while(...) I thought it would be easy to cover but i cannot figure out a way to cover it at all. removing it didn't change the outcome of any test so I deem it safe to remove.

src/Markdig.Tests/TestPipeTable.cs
```diff
@@ -101,4 +103,103 @@ public void TableWithUnbalancedCodeSpanParsesWithoutDepthLimitError()
...
+            """
+            `
+            	|| hidden text ||
+            `
+
+              | Count | Value |
+              |-------|-------|
+              | 0     | B |
+
+            """;
```
MihaZupan:
```suggestion
            """.ReplaceLineEndings("\n");
```
Tests can break depending on OS because the number of spaces in a code inline can differ based on line endings

src/Markdig.Tests/TestPipeTable.cs
```diff
@@ -101,4 +103,103 @@ public void TableWithUnbalancedCodeSpanParsesWithoutDepthLimitError()
...
+
+    [Test]
+    public void MultiLineCodeInlineWithPipeDelimitersRendersAsCode()
+    {
+        const string markdown =
+            """
+            `
+            || hidden text ||
+            `
+            """;
```
MihaZupan:
```suggestion
        string markdown =
            """
            `
            || hidden text ||
            `
            """.ReplaceLineEndings("\n");
```

src/Markdig/Extensions/Tables/PipeTableParser.cs
phillip-haydon: This was the first bit of code I put in so I guess it became unnecessary. Removed it after confirming its not affecting the outcome of things.

phillip-haydon:
Anything else remaining?

src/Markdig/Parsers/Inlines/CodeInlineParser.cs
```diff
@@ -84,15 +85,15 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
...
+                if (lookAhead[0] == '|')
                 {
-                    whitespace++;
-                }
-                if (whitespace < lookAhead.Length && lookAhead[whitespace] == '|')
-                {
-                    slice.Start = openingStart;
-                    return false;
+                    // Mirror the guard in PipeTableParser: if the next line starts with a pipe while we are inside
+                    // a table delimiter, defer to the pipe table parser so the backtick span stays intact.
```
MihaZupan:
```suggestion
                    // We saw the start of a code inline, but the close sticks are not present on the same line.
                    // If the next line starts with a pipe character, this is likely an incomplete CodeInline within a table.
                    // Treat it as regular text to avoid breaking the overall table shape.
```

src/Markdig/Parsers/Inlines/CodeInlineParser.cs
```diff
@@ -84,15 +85,15 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
                 {
                     lookAhead = lookAhead.Slice(1);
                 }
-                int whitespace = 0;
-                while (whitespace < lookAhead.Length && (lookAhead[whitespace] == ' ' || lookAhead[whitespace] == '\t'))
+                if (lookAhead[0] == '|')
```
MihaZupan, line 88: Any way this could throw?

The length check above
```c#
ReadOnlySpan<char> lookAhead = span.Length > 1 ? span.Slice(1) : ReadOnlySpan<char>.Empty;
```
isn't necessary since `span` is always non-empty when you reach that branch, but the `lookAhead[0] == '|'` could be out of range after slicing off new lines above.

phillip-haydon: I actually think the length check is unnecessary, lemme rewrite this part.

xoofx: > I actually think the length check is unnecessary, lemme rewrite this part.

Haven't checkout the code locally, but sharing the concern from @MihaZupan 

Why do you think checking length is not necessary? Looks like the previous `lookAhead = lookAhead.Slice(1);` could be at the end of the string, no?

phillip-haydon: When I debug it even tho there is nothing there taking a slice ends up with a terminator or some value. It never ends up with an out of range index which is what I would expect. I haven’t dug into why but I tried setting 1 to 10000 and it didn’t throw. Just been really busy with other stuff this week I haven’t had a chance to investigate further.

xoofx: Ok thanks, let's try to see If I have some this week-end to check this. Merging the PR.

MihaZupan: #900

Issue #886, opened 2025-08-13
Add OpenInNewWindow property to LinkInlineRenderer


danielklecha:
There is no way to force to open markdown links in a new window.
It's possible for auto links, but markdown links don't have this feature.

OpenInNewWindow property should add target=_blank if link is not image.

danielklecha:
Fixed in #894.

PR #894, opened 2025-09-23
add options for link inline

danielklecha:
Hi,

I want to fix issue #886.

I added options class but it's only used if user replace default parser.
Feel free to comment and give me clues :)

AutoLinkOptions has few more properties but I didn't implement it.

Does anything else should be changed?

### DIFF
--- a/src/Markdig/Parsers/Inlines/AutolinkOptions.cs
+++ b/src/Markdig/Parsers/Inlines/AutolinkOptions.cs
@@ -9,5 +9,5 @@ public class AutolinkOptions : LinkOptions
     /// <summary>
     /// Gets or sets a value indicating whether to enable HTML parsing. Default is <c>true</c>
     /// </summary>
-    public bool EnableHtmlParsing { get; set; }
+    public bool EnableHtmlParsing { get; set; } = true;
 }


src/Markdig/Parsers/LinkOptions.cs
```diff
@@ -0,0 +1,15 @@
+using System;
```
MihaZupan, line 5: Nit: Please copy the license header to this file as well

src/Markdig/Parsers/LinkOptions.cs
```diff
@@ -0,0 +1,15 @@
...
+using System.Threading.Tasks;
+
+namespace Markdig.Parsers.Inlines;
+
+public class LinkInlineOptions
+{
+    /// <summary>
+    /// Should the link open in a new window when clicked (false by default)
+    /// </summary>
+    public bool OpenInNewWindow { get; set; }
```
MihaZupan, line 18: We have 4 `*Link*` parsers:
1. `LinkInlineParser` - `[foo](http://foo)`
2. `AutolinkInlineParser` - `<http://foo>`
3. `AutoLinkParser` - `http://foo`
4. `JiraLinkInlineParser`

Nr. 3 and 4 already have their own options type with a `OpenInNewWindow` flag.
This PR is adding it to Nr. 1.

I think we should commit and add it for Nr. 2 as well.

Potentially could also consider a shared base class for all `LinkOptions` types and stash the `OpenInNewWindow` property there.

danielklecha:
@MihaZupan could you recheck my changes?
1. I added comment with license information
2. I created LinkOptions.cs and associated it with all four parsers
3. I moved EnableHtmlParsing into options class in AutolinkInlineParser - we break backward compatibility in this place but I think that we should keep options in one place.
4. I think that I added _target attribute correctly but please check it because I could have missed something.

danielklecha:
I fixed tests.

PR #896, opened 2025-09-27
Remove leading newline in block attributes

### LINKED ISSUE #895: Block attribute paragraph leaves leading newline in block (Pandoc spec violation)
## Bug Description
When using GenericAttributesParser, if an attribute block (e.g., `{.center}`) appears on a line by itself immediately before a block element (such as a paragraph), the resulting block element includes a leading newline (`\n`) in its content.

## Expected Behavior
According to the Pandoc Markdown specification:
> Attributes that occur immediately before a block element, on a line by themselves, affect that element.

The attribute block should be associated with the following block element, but the content of the block element should not include an extra leading newline.

## How to Reproduce
Steps:
1. Create a markdown file containing:
   ```
   {.center}
   A paragraph
   ```
2. Parse this with Markdig (with GenericAttributes enabled)
3. Observe that the paragraph block contains a leading `\n` in its content.

## Actual Behavior
The paragraph block has a leading newline before the text, which is preserved in the rendered HTML or output.

## Expected Output
The paragraph should have the attribute applied, but its content should be:
```
A paragraph
```
without any leading newline.

## Reference
Pandoc Markdown spec: https://pandoc.org/MANUAL.html#extension-attributes

## Additional Context
This may be due to how the parser transitions from the attribute block to the following block, and does not trim the whitespace. This leads to non-compliance with Markdown attribute extension behavior as described in Pandoc.

Please consider aligning this behavior with the Pandoc spec by trimming/removing the leading newline when attaching attributes from an attribute-only paragraph to the next block.

Asttear:
fix #895

### DIFF
--- a/src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
+++ b/src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
@@ -109,6 +109,10 @@ public class GenericAttributesParser : InlineParser
             {
                 isValid = true;
                 line.SkipChar(); // skip }
+                if (line.CurrentChar == '\n')
+                {
+                    line.SkipChar(); // skip \n
+                }
                 break;
             }
 
--- a/src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
+++ b/src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
@@ -109,9 +109,14 @@ public class GenericAttributesParser : InlineParser
             {
                 isValid = true;
                 line.SkipChar(); // skip }
+                // skip line breaks
                 if (line.CurrentChar == '\n')
                 {
-                    line.SkipChar(); // skip \n
+                    line.SkipChar();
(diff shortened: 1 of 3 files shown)


src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
```diff
@@ -109,6 +109,10 @@ public static bool TryParse(ref StringSlice slice, [NotNullWhen(true)] out HtmlA
             {
                 isValid = true;
                 line.SkipChar(); // skip }
+                if (line.CurrentChar == '\n')
```
xoofx, line 113: This should also handle `\r` and `\n` as in other places in the codebase.

```c#
if (line.CurrentChar == '\n')
{
    line.SkipChar();
} 
else if (line.CurrentChar == '\r')
{
    line.SkipChar();
    if (line.PeekChar() == '\n')
    {
        line.SkipChar();
    }
}
```

PR #898, opened 2025-10-10
Add HTML tag filtering extension with whitelist and blacklist options

### LINKED ISSUE #400: Best way to have an allow/block list of HTML elements/attributes?
Not really suggesting this as a base feature (although that would be great), more considering writing it as an extension myself and curious if someone on the project would have guidance on the best way to go about it (what part of the pipeline to engage with, etc). We are using markdig as the markdown parser behind https://docs.microsoft.com and would like to consider sanitizing our input more robustly. Thanks!

Shane32:
Closes #400

### DIFF
--- a/src/Markdig/Extensions/HtmlTagFilter/FilteredAutolinkInlineParser.cs
+++ b/src/Markdig/Extensions/HtmlTagFilter/FilteredAutolinkInlineParser.cs
@@ -118,7 +118,7 @@ public class FilteredAutolinkInlineParser : AutolinkInlineParser
             // Check if the next character suggests this is an HTML tag
             // HTML tags have whitespace, '>', '/', or attributes after the tag name
             // Autolinks have ':' or '@' 
-            if (c == ':' || c == '@')
+            if (c is ':' or '@')
             {
                 return false; // This is likely an autolink, not an HTML tag
             }


src/Markdig/Extensions/HtmlTagFilter/FilteredAutolinkInlineParser.cs
```diff
@@ -0,0 +1,136 @@
...
+            {
+                return false;
+            }
+            slice.SkipChar();
+
+            var c = slice.CurrentChar;
+
+            // Check if this looks like an autolink (not an HTML tag)
+            // Autolinks don't start with /, !, or ?
+            if (c == '/' || c == '!' || c == '?')
```
gfoidl:
```suggestion
            if (c is '/' or '!' or '?')
```
At the moment almost identical code is emitted.
In the future Roslyn may generate an optimized test for this pattern match. So if Roslyn can do this, we get a free perf-win then.

src/Markdig/Extensions/HtmlTagFilter/FilteredAutolinkInlineParser.cs
```diff
@@ -0,0 +1,136 @@
...
+
+            if (tagLength == 0)
+            {
+                return false;
+            }
+
+            // Check if the next character suggests this is an HTML tag
+            // HTML tags have whitespace, '>', '/', or attributes after the tag name
+            // Autolinks have ':' or '@' 
+            if (c == ':' || c == '@')
```
gfoidl:
```suggestion
            if (c is ':' or '@')
```
Note: there's no common set, so Roslyn probably won't optimize this in the future anyway. But for consistency with the above suggestion it should be written this way.

MihaZupan:
I take it the intention here is primarily to rely on this validation when dealing with untrusted inputs?

I'm concerned that this would give users false confidence in Markdig providing security guarantees which don't hold up in practice.
E.g. while we have some basic mitigations against malformed inputs in place, we make no guarantees about linearish parsing complexity either. There are other libraries that should be doing a more thorough job of XSS mitigation.

xoofx:
> I'm concerned that this would give users false confidence in Markdig providing security guarantees which don't hold up in practice.

I share similar concern, because it would open even more work for sanitizing attributes (which is where you can have even more problems) and markdig should not be a place to do such work, as there are more robust and specialized approaches. For example, using [HtmlSanitizer](https://github.com/mganss/HtmlSanitizer) or similar to filter HTML output from markdig.

@Shane32 could you elaborate what is the usecase of this PR that could not be covered by HtmlSanitizer?

Shane32:
I'm not familiar with HtmlSanitizer, and I don't use Markdig.  I'm trying to help Microsoft solve https://github.com/NuGet/NuGetGallery/issues/8644 , which will be better for the entire community.  This was a suggested method, and you have an open issue here; hence I posted a PR.

Shane32:
Considering that scripts may be contained within attributes, and that the primary use case is to remove scripts, perhaps this is a poor approach.  Logically, it would seem that it would make sense for a markdown conversion tool to allow some basic HTML.  Perhaps this PR could be changed to only support whitelisted tags with no attributes.  (According to ChatGPT, there are potentially risks with unsanitized CSS, even if minimal.)  Eliminate the blacklisting capability given it would be prone to insufficient blacklists.  And this would allow `<br/>` and `<pre>` and `<sup>` and a few others that would be useful.  Unfortunately it wouldn't help with `<font>` and some other common tags.

xoofx:
> I'm not familiar with HtmlSanitizer, and I don't use Markdig. I'm trying to help Microsoft solve [NuGet/NuGetGallery#8644](https://github.com/NuGet/NuGetGallery/issues/8644) , which will be better for the entire community. This was a suggested method, and you have an open issue here; hence I posted a PR.


I commented on the issue here https://github.com/NuGet/NuGetGallery/issues/8644#issuecomment-1020943480 and my recommendation at this time still stands ☺️ 

I'm going to close this PR and not merge it. My apology, as I don't think it is the right solution for the problem, and it might confuse people using it or sets false expectations. Thank you for trying to provide a fix and resolve this issue!

Sanitizing HTML output of markdig is pretty easy to do with an HTML sanitizer, so NuGet gallery has probably another problem (fear of sanitizing issues...etc.) that they will have to address.

Shane32:
No problem. Thanks!

PR #903, opened 2025-10-16
Add a basic fuzzing project

MihaZupan:
(On Windows)
As simple as running `.\run-fuzzer.ps1` from `src/Markdig.Fuzzing`.

### DIFF
--- a/src/Markdig.Fuzzing/Markdig.Fuzzing.csproj
+++ b/src/Markdig.Fuzzing/Markdig.Fuzzing.csproj
@@ -5,6 +5,7 @@
     <TargetFramework>net9.0</TargetFramework>
     <ImplicitUsings>enable</ImplicitUsings>
     <Nullable>enable</Nullable>
+    <IsPackable>false</IsPackable>
   </PropertyGroup>
 
   <ItemGroup>


src/Markdig.Fuzzing/run-fuzzer.ps1
```diff
@@ -0,0 +1,86 @@
...
+
+$libFuzzer = "libfuzzer-dotnet-windows.exe"
+$outputDir = "bin"
+
+function Get-LibFuzzer {
+    param (
+        [string]$Path
+    )
+
+    $libFuzzerUrl = "https://github.com/Metalnem/libfuzzer-dotnet/releases/download/v2025.05.02.0904/libfuzzer-dotnet-windows.exe"
```
AmadeusW, line 15: that doesn't seem safe, you're downloading a binary from the internet and feeding it contents of your directory.

MihaZupan: But we're also pinning the version to a known build (via the hash), and the author is a MS FTE. Did you have an alternative approach in mind?

FWIW we're doing [the same thing](https://github.com/dotnet/runtime/blob/8f0e8abd5c5a315de2fb7e8327f1c61d5b8f3dbd/src/libraries/Fuzzing/DotnetFuzzing/Program.cs#L108-L111) for fuzzing .NET libraries in dotnet/runtime.

(we're similarly pulling down the sharpfuzz CLI tool)

xoofx: Agreed, the hash makes it safe enough.

elmigranto: I find it very spooky as well.

> Agreed, the hash makes it safe enough.

We don't really know what's under current hash, right? Could be `if (year is 2026) run_virus()`.

MihaZupan:
It's not happening as part of build, only if you run the `.\run-fuzzer.ps1` script yourself.
(The libfuzzer executable & the CLI tool that is. We are referencing the `SharpFuzz` NuGet package as part of the build)

> I find it a little spooky to be downloading and running third-party executables as part of build.

FWIW referencing any NuGet package is effectively this.

elmigranto:
> FWIW referencing any NuGet package is effectively this.

Yes, that is spooky as well :)

PR #905, opened 2025-10-20
Fix pipe table parsing with a leading paragraph

### LINKED ISSUE #818: Tables are not rendered properly unless there is a new line before them

MihaZupan:
Closes #818
Addresses issues referenced in https://github.com/xoofx/markdig/pull/885#issuecomment-3419948799, https://github.com/xoofx/markdig/pull/885#issuecomment-3419969194

This PR effectively reverts #885 as that change was swallowing whatever information was there in the paragraph before the table.
E.g. for
```md
Some text
| A |
|---|
| B |
```
"Some text" would be lost.

Instead I relaxed the table validation to no longer enforce starting on the first line (`localLineIndex > 0` and `deltaLine > 0` checks), and also save the current block in the AST before we replace it with a table.

Given this is happening while we're parsing inlines, I don't think we can easily insert an extra block at the current layer, so I instead added the leading paragrah as the first child of the table, and the first element is special-cased in the renderer.
That is, instead of
```
document
  paragraph
    some text
  table
    A ...
```
we're instead producing
```
document
  table
    paragraph
      some text
    A ...
```

Since `TableRow` is a public type, this does risk breaking code that was walking the AST and assuming that all children of a `Table` are `TableRow`s - this seems plausible given our own renderer did that.
I'm open to alternative suggestions.

### DIFF
--- a/src/Markdig/Extensions/Tables/HtmlTableRenderer.cs
+++ b/src/Markdig/Extensions/Tables/HtmlTableRenderer.cs
@@ -2,12 +2,10 @@
 // This file is licensed under the BSD-Clause 2 license.
 // See the license.txt file in the project root for more information.
 
-using System.Diagnostics;
 using System.Globalization;
 
 using Markdig.Renderers;
 using Markdig.Renderers.Html;
-using Markdig.Syntax;
 
 namespace Markdig.Extensions.Tables;
 
@@ -19,12 +17,6 @@ public class HtmlTableRenderer : HtmlObjectRenderer<Table>
 {
     protected override void Write(HtmlRenderer renderer, Table table)
     {
-        // If the table started with a leading paragraph, that block will be stored as the first child.
-        if (table.Count > 0 && table[0] is not TableRow)
-        {
-            renderer.Write(table[0]);
-        }
-
         if (renderer.EnableHtmlForBlock)
         {
             renderer.EnsureLine();
@@ -55,14 +47,9 @@ public class HtmlTableRenderer : HtmlObjectRenderer<Table>
                 }
             }
 
-            foreach (Block rowObj in table)
+            foreach (var rowObj in table)
             {
-                if (rowObj is not TableRow row)
-                {
-                    Debug.Assert(ReferenceEquals(rowObj, table[0]));
-                    continue;
-                }
-
+                var row = (TableRow)rowObj;
                 if (row.IsHeader)
                 {
                     // Allow a single thead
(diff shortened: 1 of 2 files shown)


MihaZupan:
Alternative approach that'd get us the expected syntax tree shape and avoids modifying the renderer: https://github.com/xoofx/markdig/compare/master...MihaZupan:table-para2
This being a hack that says "insert the block, but skip processing it in `MarkdownParser.ProcessInlines`".

Could theoretically break someone that wanted to look at the table shape from a different parser (since the table isn't inserted until later), but that seems less likely?

xoofx:
> Could theoretically break someone that wanted to look at the table shape from a different parser (since the table isn't inserted until later), but that seems less likely?

Yes, maybe that workaround is ok. Table parsing is a bit awful unfortunately, so this new code doesn't change that 😅

RickStrahl:
Can the same logic that's used for GridTable parsing to find the start of the block? GridTables seem to have the correct behavior.

ps. I haven't had time to look at the code beyond what was posted here, just going on the rendering behavior in the current 0.42 release.

MihaZupan:
Maybe.
Grid tables are parsed with a block parser, while the pipe table is parsed with an inline one that effectively hijacks a paragraph block and replaces it with a table block.

I haven't been bored enough yet to look at what it'd take to turn it into a block parser instead 😄
Maybe Alexandre remembers more details.

RickStrahl:
Yeah I know what you mean - I took a look at the code a few days ago thinking I could help in a hurry, but without setting aside a good chunk of time to understand the flow these block parsers are easy to mess up. I made a custom fix a few years ago and discovered months later it was breaking in completely unexpected ways.

Thanks for your effort here - really appreciate you taking the time!

Aloha.

MihaZupan:
Note that this change makes it work when you have leading text before the table.
If you have text right after the table, that still won't work.

Issue #910, opened 2025-10-29
German Umlaut handling in urilize


mos379:
When generating proper Ids for headers I ran into the problem that the German Umlauts are not handled properly and this causes some problems

e.g. 
`<h2 id="-revolution-re-funktionen">🚀 Revolutionäre Funktionen</h2>`
while 
`<h2 id="-revolutionaere-funktionen">🚀 Revolutionäre Funktionen</h2>`
would be appropriate. 

since the IAutoIdentifierGenerator is not public its not possible to overwrite, so to include this in the implementation of the standard would be great.
so for example in the LinkHelper

[Urilize.txt](https://github.com/user-attachments/files/23208116/Urilize.txt)

xoofx:
Interesting, could you open a proper [Pull-Request](https://github.com/xoofx/markdig/pulls)?

I'm curious also, is it really only for the German language, or there is something that we are missing with the .NET implementation regarding such transform?

MihaZupan:
Can you share the code you're using?

This is the current behavior I see
```c#
AutoLink
<h1 id="revolutionäre-funktionen">🚀 Revolutionäre Funktionen</h1>

AllowOnlyAscii, AutoLink (default)
<h1 id="revolutionare-funktionen">🚀 Revolutionäre Funktionen</h1>

GitHub
<h1 id="-revolutionäre-funktionen">🚀 Revolutionäre Funktionen</h1>

AllowOnlyAscii, GitHub -- the GitHub format doesn't pay attention to AllowOnlyAscii rn
<h1 id="-revolutionäre-funktionen">🚀 Revolutionäre Funktionen</h1>
```

with the mapping to ASCII defined here for AutoLink:
https://github.com/xoofx/markdig/blob/8c01cf054971a1ec4cd663edaca6e2d035236133/src/Markdig/Helpers/CharNormalizer.cs#L67
or just `char.ToLowerInvariant` for GitHub.

mos379:
> Interesting, could you open a proper [Pull-Request](https://github.com/xoofx/markdig/pulls)?
> 
> I'm curious also, is it really only for the German language, or there is something that we are missing with the .NET implementation regarding such transform?

Yes, there is a standard .NET approach, but it has an important limitation for German umlauts. 
[...]

PR #911, opened 2025-10-29
feat(link-helper): improve ASCII normalization handling

mos379:
feat(link-helper): improve ASCII normalization handling

Enhanced the `Urilize` method to better handle ASCII normalization and special characters. Added support for decomposing characters when `allowOnlyAscii` is true and skipping diacritical marks. Introduced handling for special German, Scandinavian, and Icelandic characters via new helper methods: `IsSpecialScandinavianOrGermanChar` and `NormalizeScandinavianOrGermanChar`.

Reorganized `using` directives for better clarity. Updated the processing loop in `Urilize` to handle normalized spans and ASCII equivalents more effectively. These changes improve link generation compatibility across various languages.
#910

### DIFF
--- a/src/Markdig.Tests/TestLinkHelper.cs
+++ b/src/Markdig.Tests/TestLinkHelper.cs
@@ -327,8 +327,8 @@ public class TestLinkHelper
         Assert.AreEqual(expectedResult, LinkHelper.Urilize(input, true));
     }
 
-    [TestCase("bær", "br")]
-    [TestCase("bør", "br")]
+    [TestCase("bær", "baer")]
+    [TestCase("bør", "boer")]
     [TestCase("bΘr", "br")]
     [TestCase("四五", "")]
     public void TestUrilizeOnlyAscii_NonAscii(string input, string expectedResult)


src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -101,6 +130,50 @@ public static string Urilize(ReadOnlySpan<char> headingText, bool allowOnlyAscii
...
+               c == 'Þ' || c == 'Ð';
+    }
+
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    private static ReadOnlySpan<char> NormalizeScandinavianOrGermanChar(char c)
+    {
+        return c switch
+        {
+            // German
+            'ä' => "ae",
```
xoofx, line 154: I'm curious why introducing this code and not changing directly https://github.com/xoofx/markdig/blob/8c01cf054971a1ec4cd663edaca6e2d035236133/src/Markdig/Helpers/CharNormalizer.cs#L67 could you elaborate?

mos379: Yes I did not want to change the ASCII routine as this as far as I understood is not used in the GitHub variation.
In German the proper replacement is always the two letter version, not the single letter.

xoofx:
Sorry to come back on this only now. Some tests are failing, if you could have a look? Thanks!

mos379:
Gotta fix the tests, working on it.
Actually I'll see what Copilot can do :D

mos379:
Great Thanks!

Issue #890, opened 2025-09-14
Support CJK Friendly Emphasis extension


tats-u:
CommonMark's emphasis has a specification bug that is incompatible with Chinese and Japanese.

<div lang="ja">

> **この文は太字になりません。**この文のせいで。

</div>

> __This sentence will not be bolded.__ It is due to this sentence.

<img width="788" height="151" alt="Image" src="https://github.com/user-attachments/assets/92be4841-99b1-4b1d-9e32-50abee477997" />

https://github.com/tats-u/markdown-cjk-friendly

This extension significantly improves compatibility with Chinese and Japanese. `**` in all the above examples will be properly recognized as `<strong>` thanks to this extension.

Supported (opt-_out_):

- VitePress

Supported (opt-in):

- Comrak / commonmarker

Blocked by #860

PR #913, opened 2025-11-13
Recognize supplementary characters

### LINKED ISSUE #860: Recognize supplementary punctuations around emphasis delimiter runs
https://github.com/xoofx/markdig/blob/fb3fe8b261dfdec82edc2d1b07735931df04bbd4/src/Markdig/Helpers/CharHelper.cs#L37-L60

- https://github.com/commonmark/commonmark.js/pull/297

The specs says "A [character](https://spec.commonmark.org/0.31.2/#character) is a Unicode code point." However, the `char` type is just a UTF-16 code *unit*, which does not cover supplementary characters (whose code points are U+100000 or greater).

https://spec.commonmark.org/0.31.2/#character

tats-u:
Fixes #860

### DIFF
--- a/src/Markdig.Tests/TestStringSlice.cs
+++ b/src/Markdig.Tests/TestStringSlice.cs
@@ -3,131 +3,163 @@
 // See the license.txt file in the project root for more information.
 
 using Markdig.Helpers;
-using System;
-using System.Collections.Generic;
-using System.Drawing;
-using System.Linq;
-#if TEST_NETSTANDARD_BUILD_IN_MODERN_NET
-using Markdig.Polyfills.System.Text;
-#else
-using System.Text;
-#endif
-using System.Threading.Tasks;
 
 namespace Markdig.Tests;
 
 [TestFixture]
-internal class TestStringSlice
+public class TestStringSlice
 {
-    // TODO: Add more tests for StringSlice
-
-    // Old versions of modern .NET (that meets NETCOREAPP but not the following #if) try to load the polyfilled Rune in .NET Standard 2.x build of Markdig while they have built-in Rune too
-    // Adjust this condition to match the minimal modern (not .NET Standard or Framework) .NET target framework of Markdig (not Markdig.Tests)
-#if NET8_0_OR_GREATER
-
+#if NET
     [Test]
     public void TestRuneBmp()
     {
         var slice = new StringSlice("01234");
 
-        Assert.AreEqual((Rune)'0', slice.CurrentRune);
-        Assert.AreEqual((Rune)'1', slice.NextRune());
-        Assert.AreEqual((Rune)'2', slice.NextRune());
-        Assert.AreEqual((Rune)'2', slice.CurrentRune);
+        Assert.AreEqual('0', slice.CurrentRune.Value);
+        Assert.AreEqual(0, slice.Start);
+        Assert.AreEqual('1', slice.NextRune().Value);
+        Assert.AreEqual(1, slice.Start);
+        Assert.AreEqual('2', slice.NextRune().Value);
+        Assert.AreEqual(2, slice.Start);
+        Assert.AreEqual('2', slice.CurrentRune.Value);
         Assert.AreEqual("234", slice.ToString());
-        Assert.AreEqual((Rune)'3', slice.PeekRuneExtra(1));
-        Assert.AreEqual((Rune)'4', slice.PeekRuneExtra(2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(3));
-        Assert.AreEqual((Rune)'1', slice.PeekRuneExtra(-1));
-        Assert.AreEqual((Rune)'0', slice.PeekRuneExtra(-2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-3));
-        Assert.AreEqual((Rune)'0', slice.RuneAt(0));
-        Assert.AreEqual((Rune)'1', slice.RuneAt(1));
-        Assert.AreEqual((Rune)'2', slice.RuneAt(2));
-        Assert.AreEqual((Rune)'3', slice.RuneAt(3));
-        Assert.AreEqual((Rune)'4', slice.RuneAt(4));
+        Assert.AreEqual('3', slice.PeekRuneExtra(1).Value);
+        Assert.AreEqual('4', slice.PeekRuneExtra(2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(3).Value);
+        Assert.AreEqual('1', slice.PeekRuneExtra(-1).Value);
+        Assert.AreEqual('0', slice.PeekRuneExtra(-2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-3).Value);
+        Assert.AreEqual('0', slice.RuneAt(0).Value);
+        Assert.AreEqual('1', slice.RuneAt(1).Value);
+        Assert.AreEqual('2', slice.RuneAt(2).Value);
+        Assert.AreEqual('3', slice.RuneAt(3).Value);
+        Assert.AreEqual('4', slice.RuneAt(4).Value);
+        Assert.AreEqual(2, slice.Start);
     }
 
     [Test]
     public void TestRuneSupplementaryOnly()
     {
         var slice = new StringSlice("𝟎𝟏𝟐𝟑𝟒");
+        Assert.AreEqual(10, slice.Length);
 
         // 𝟎 = U+1D7CE, 𝟐 = U+1D7D0
-        Assert.AreEqual((Rune)0x1D7CE, slice.CurrentRune); // 𝟎
-        Assert.AreEqual((Rune)0x1D7CF, slice.NextRune()); // 𝟏
-        Assert.AreEqual((Rune)0x1D7D0, slice.NextRune()); // 𝟐
-        Assert.AreEqual((Rune)0x1D7D0, slice.CurrentRune); // 𝟐
+        Assert.AreEqual(0x1D7CE, slice.CurrentRune.Value); // 𝟎
+        Assert.AreEqual(0, slice.Start);
+        Assert.AreEqual(0x1D7CF, slice.NextRune().Value); // 𝟏
+        Assert.AreEqual(2, slice.Start);
+        Assert.AreEqual(0x1D7D0, slice.NextRune().Value); // 𝟐
+        Assert.AreEqual(4, slice.Start);
+        Assert.AreEqual(0x1D7D0, slice.CurrentRune.Value); // 𝟐
         Assert.AreEqual("𝟐𝟑𝟒", slice.ToString());
         // CurrentRune occupies 2 `char`s, so next Rune starts at index 2
-        Assert.AreEqual((Rune)0x1D7D1, slice.PeekRuneExtra(2)); // 𝟑
-        Assert.AreEqual((Rune)0x1D7D2, slice.PeekRuneExtra(4)); // 𝟒
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(6));
-        Assert.AreEqual((Rune)0x1D7CF, slice.PeekRuneExtra(-1)); // 𝟏
-        Assert.AreEqual((Rune)0x1D7CE, slice.PeekRuneExtra(-3)); // 𝟎
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-5));
-        Assert.AreEqual((Rune)0x1D7CE, slice.RuneAt(0)); // 𝟎
-        Assert.AreEqual((Rune)0x1D7CF, slice.RuneAt(2)); // 𝟏
-        Assert.AreEqual((Rune)0x1D7D0, slice.RuneAt(4)); // 𝟐
-        Assert.AreEqual((Rune)0x1D7D1, slice.RuneAt(6)); // 𝟑
-        Assert.AreEqual((Rune)0x1D7D2, slice.RuneAt(8)); // 𝟒
+        Assert.AreEqual(0x1D7D1, slice.PeekRuneExtra(2).Value); // 𝟑
+        Assert.AreEqual(0x1D7D2, slice.PeekRuneExtra(4).Value); // 𝟒
+        Assert.AreEqual(0, slice.PeekRuneExtra(6).Value);
+        Assert.AreEqual(0x1D7CF, slice.PeekRuneExtra(-1).Value); // 𝟏
+        Assert.AreEqual(0x1D7CE, slice.PeekRuneExtra(-3).Value); // 𝟎
+        Assert.AreEqual(0, slice.PeekRuneExtra(-5).Value);
+        Assert.AreEqual(0x1D7CE, slice.RuneAt(0).Value); // 𝟎
+        Assert.AreEqual(0x1D7CF, slice.RuneAt(2).Value); // 𝟏
+        Assert.AreEqual(0x1D7D0, slice.RuneAt(4).Value); // 𝟐
+        Assert.AreEqual(0x1D7D1, slice.RuneAt(6).Value); // 𝟑
+        Assert.AreEqual(0x1D7D2, slice.RuneAt(8).Value); // 𝟒
         // The following usages are not expected. You should take into consideration the `char`s that the Rune you just acquired occupies.
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-4));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(1));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(3));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(5));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(1));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(3));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(5));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(7));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(9));
+        Assert.AreEqual(0, slice.PeekRuneExtra(-4).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(1).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(3).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(5).Value);
+        Assert.AreEqual(0, slice.RuneAt(1).Value);
+        Assert.AreEqual(0, slice.RuneAt(3).Value);
+        Assert.AreEqual(0, slice.RuneAt(5).Value);
+        Assert.AreEqual(0, slice.RuneAt(7).Value);
+        Assert.AreEqual(0, slice.RuneAt(9).Value);
+        Assert.AreEqual(4, slice.Start);
     }
 
     [Test]
     public void TestRuneIsolatedHighSurrogate()
     {
         var slice = new StringSlice("\ud800\ud801\ud802\ud803\ud804");
-        Assert.AreEqual((Rune)'\0', slice.CurrentRune);
-        Assert.AreEqual((Rune)'\0', slice.NextRune());
+        Assert.AreEqual(0, slice.CurrentRune.Value);
+        Assert.AreEqual(0, slice.Start);
+        Assert.AreEqual(0, slice.NextRune().Value);
+        Assert.AreEqual(0, slice.CurrentRune.Value);
         Assert.AreEqual('\ud801', slice.CurrentChar);
-        Assert.AreEqual((Rune)'\0', slice.NextRune());
+        Assert.AreEqual(1, slice.Start);
+        Assert.AreEqual(0, slice.NextRune().Value);
+        Assert.AreEqual(2, slice.Start);
         Assert.AreEqual('\ud802', slice.CurrentChar);
-        Assert.AreEqual((Rune)'\0', slice.CurrentRune);
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-3));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-1));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(1));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(3));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(0));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(1));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(2));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(3));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(4));
+        Assert.AreEqual(0, slice.CurrentRune.Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-3).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-1).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(1).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(3).Value);
+        Assert.AreEqual(0, slice.RuneAt(0).Value);
+        Assert.AreEqual(0, slice.RuneAt(1).Value);
+        Assert.AreEqual(0, slice.RuneAt(2).Value);
+        Assert.AreEqual(0, slice.RuneAt(3).Value);
+        Assert.AreEqual(0, slice.RuneAt(4).Value);
+        Assert.AreEqual(2, slice.Start);
     }
 
     [Test]
     public void TestRuneIsolatedLowSurrogate()
     {
         var slice = new StringSlice("\udc00\udc01\udc02\udc03\udc04");
-        Assert.AreEqual((Rune)'\0', slice.CurrentRune);
-        Assert.AreEqual((Rune)'\0', slice.NextRune());
+        Assert.AreEqual(0, slice.CurrentRune.Value);
+        Assert.AreEqual(0, slice.NextRune().Value);
         Assert.AreEqual('\udc01', slice.CurrentChar);
-        Assert.AreEqual((Rune)'\0', slice.NextRune());
+        Assert.AreEqual(0, slice.NextRune().Value);
         Assert.AreEqual('\udc02', slice.CurrentChar);
-        Assert.AreEqual((Rune)'\0', slice.CurrentRune);
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-3));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(-1));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(1));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(2));
-        Assert.AreEqual((Rune)'\0', slice.PeekRuneExtra(3));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(0));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(1));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(2));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(3));
-        Assert.AreEqual((Rune)'\0', slice.RuneAt(4));
+        Assert.AreEqual(0, slice.CurrentRune.Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-3).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(-1).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(1).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(2).Value);
+        Assert.AreEqual(0, slice.PeekRuneExtra(3).Value);
+        Assert.AreEqual(0, slice.RuneAt(0).Value);
+        Assert.AreEqual(0, slice.RuneAt(1).Value);
+        Assert.AreEqual(0, slice.RuneAt(2).Value);
+        Assert.AreEqual(0, slice.RuneAt(3).Value);
+        Assert.AreEqual(0, slice.RuneAt(4).Value);
+    }
+
+    [Test]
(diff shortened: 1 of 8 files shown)


src/Markdig/Helpers/CharHelper.cs
```diff
@@ -199,6 +200,9 @@ public static bool IsWhitespace(this char c)
         return IsWhitespaceRare(c);
     }
 
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    public static bool IsWhitespace(this Rune r) => r.IsBmp && IsWhitespace((char)r.Value);
```
MihaZupan: Please make these new helpers internal.

No two specs agree on what space means, so I'd avoid exposing more ambiguous extensions on common types (it's not obvious this is CommonMark-specific).

(same for all new helpers in this file)

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -263,6 +267,52 @@ public static void CheckUnicodeCategory(this char c, out bool space, out bool pu
         }
     }
 
+#if !(NETSTANDARD2_1_OR_GREATER || NETCOREAPP2_1_OR_GREATER)
+    private static Lazy<Func<int, UnicodeCategory>?> GetUnicodeCategoryReflection =
```
MihaZupan: I think we should fallback to `CharUnicodeInfo.GetUnicodeCategory(value.ToString(), 0)` instead.
It's the same thing that .NET itself uses in its Rune impl for Framework.

https://github.com/dotnet/runtime/blob/d594a04c1a2426ef295448c756f55965dedfc044/src/libraries/System.Private.CoreLib/src/System/Text/Rune.cs#L1279-L1287

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -125,6 +126,31 @@ public readonly char CurrentChar
...
+            {
+                if (start + 1 > End) return default;
+                var second = Text[start + 1];
+                if (!char.IsLowSurrogate(second)) return default;
+                return new Rune(first, second);
+            }
+            if (start < 1) return default;
+            var trueFirst = Text[start - 1];
+            if (!char.IsHighSurrogate(trueFirst)) return default;
+            return new Rune(trueFirst, first);
```
MihaZupan: This seems like a usage error, I would avoid reading backwards in these helpers.
(same for all of these)

src/Markdig/Markdig.targets
```diff
@@ -28,6 +28,10 @@
     <PackageReference Include="System.Memory" Version="4.6.0" />
   </ItemGroup>
 
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net462' OR '$(TargetFramework)' == 'netstandard2.0' OR '$(TargetFramework)' == 'netstandard2.1'">
+    <PackageReference Include="Shim.System.Text.Rune" Version="6.0.2" />
+  </ItemGroup>
```
MihaZupan: Please remove the new dependency, we should polyfill the required Rune logic instead
https://github.com/xoofx/markdig/tree/master/src/Markdig/Polyfills

The new `Rune` helpers on `StringSlice` can be internal (or at least internal on targets without `Rune` in the base libraries).

src/Markdig/Helpers/CharHelper.cs
tats-u: > No two specs agree on what space means

https://spec.commonmark.org/0.31.2/#space says:

> A [space](https://spec.commonmark.org/0.31.2/#space) is U+0020.

c.f.:

> A [Unicode whitespace character](https://spec.commonmark.org/0.31.2/#unicode-whitespace-character) is a character in the Unicode Zs general category, or a tab (U+0009), line feed (U+000A), form feed (U+000C), or carriage return (U+000D).
>
> [Unicode whitespace](https://spec.commonmark.org/0.31.2/#unicode-whitespace) is a sequence of one or more [Unicode whitespace characters](https://spec.commonmark.org/0.31.2/#unicode-whitespace-character).

We may have to change comments or method names because Markdig sometimes seems to use the term "space" to mean "Unicode whitespace character" in CommonMark spec.

tats-u: I added XML comments that refer to the spec page and terms.

src/Markdig/Helpers/StringSlice.cs
tats-u: I probably deleted it because it was probably not used.

tats-u:
I ran Unit Test in my VS 2026 with .NET 6/8/9 locally and all test cases succeeded.
Sorry for having submitted the half-baked PR.

src/Markdig/Extensions/SmartyPants/SmartyPantsInlineParser.cs
```diff
@@ -36,7 +36,7 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
         // -- 	– 	&ndash; 	'ndash'
         // --- 	— 	&mdash; 	'mdash'
 
-        var pc = slice.PeekCharExtra(-1);
+        var pc = slice.PeekRuneExtra(-1);
         var c = slice.CurrentChar;
```
MihaZupan: Since `c` isn't reused anymore, let's remove the extra local and use the openingChar in the switch

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -69,10 +70,10 @@ public static class CharHelper
     private static readonly SearchValues<char> s_escapableSymbolChars = SearchValues.Create("!\"#$%&'()*+,-./:;<=>?@[\\]^_`{|}~•");
 
     [MethodImpl(MethodImplOptions.AggressiveInlining)]
-    private static bool IsPunctuationException(char c) =>
-        c is '−' or '-' or '†' or '‡';
+    private static bool IsPunctuationException(Rune c) =>
+        c.IsBmp && (char)c.Value is '−' or '-' or '†' or '‡';
 
-    public static void CheckOpenCloseDelimiter(char pc, char c, bool enableWithinWord, out bool canOpen, out bool canClose)
+    internal static void CheckOpenCloseDelimiter(Rune pc, Rune c, bool enableWithinWord, out bool canOpen, out bool canClose)
```
MihaZupan: The char-based overload should continue to exist and be public

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -180,6 +181,11 @@ public static bool IsAcrossTab(int column)
         return (column & (TabSize - 1)) != 0;
     }
 
+    /// <summary>
+    /// <see langword="true"/> if the character is an <see href="https://spec.commonmark.org/0.31.2/#unicode-whitespace-character">Unicode whitespace character</see>.
+    /// </summary>
+    /// <param name="c">The character or surrogate code unit to evaluate.</param>
```
MihaZupan:
```suggestion
    /// <param name="c">The character to evaluate.</param>
```

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -243,7 +252,13 @@ public static bool IsEscapableSymbol(this char c)
...
     }
 
-    // Check if a char is a space or a punctuation
+    /// <summary>
+    /// Checks the Unicode category of the given character and determines whether it is a whitespace or punctuation character.
+    /// </summary>
+    /// <param name="c">The character to check.</param>
+    /// <param name="space">Output parameter indicating whether the character is a whitespace character.</param>
+    /// <param name="punctuation">Output parameter indicating whether the character is a punctuation character.</param>
+
```
MihaZupan:
```suggestion
```

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -306,22 +361,37 @@ public static bool IsZero(this char c)
         return c == '\0';
     }
 
+    /// <summary>
+    /// Returns <see langword="true"/> if the character is a <see href="https://spec.commonmark.org/0.31.2/#space">space</see> (U+0020).
+    /// </summary>
+    /// <param name="c">The character or surrogate code unit to evaluate</param>
```
MihaZupan:
```suggestion
    /// <param name="c">The character to evaluate</param>
```

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -306,22 +361,37 @@ public static bool IsZero(this char c)
...
-        // 2.1 Characters and lines 
+        // 2.1 Characters and lines
         // A space is U+0020.
         return c == ' ';
     }
 
+    /// <summary>
+    /// Returns <see langword="true"/> if the character is a <see href="https://spec.commonmark.org/0.31.2/#tab">tab</see> (U+0009).
+    /// </summary>
+    /// <param name="c">The character or surrogate code unit to evaluate</param>
```
MihaZupan:
```suggestion
    /// <param name="c">The character to evaluate</param>
```

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -306,22 +361,37 @@ public static bool IsZero(this char c)
...
-        // 2.1 Characters and lines 
+        // 2.1 Characters and lines
         // A space is U+0009.
         return c == '\t';
     }
 
+    /// <summary>
+    /// Returns <see langword="true"/> if the character is a <see href="https://spec.commonmark.org/0.31.2/#space">space</see> (U+0020) or <see href="https://spec.commonmark.org/0.31.2/#tab">tab</see> (U+0009).
+    /// </summary>
+    /// <param name="c">The character or surrogate code unit to evaluate.</param>
```
MihaZupan:
```suggestion
    /// <param name="c">The character to evaluate.</param>
```

xoofx:
Thanks for the work! LGTM. Quite heavy with the Rune polyfill but we don't have much choices.

@MihaZupan anything other feedback?

tats-u:
Rune is based on the latest snapshot of https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Text/Rune.cs,a0cdde85f676b935.

tats-u:
If we adopt C# 14, we can use [Extension Members](https://learn.microsoft.com/dotnet/csharp/whats-new/csharp-14#extension-members) instead of creating the `UnicodeUtilitySupplement` class.

src/Markdig/Parsers/Inlines/EmphasisInlineParser.cs
```diff
@@ -179,12 +180,12 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
...
         }
 
-        char c = slice.CurrentChar;
+        Rune c = slice.CurrentRune;
 
         // The following character is actually an entity, we need to decode it
         if (HtmlEntityParser.TryParse(ref slice, out string? htmlString, out int htmlLength))
         {
-            c = htmlString[0];
+            Rune.DecodeFromUtf16(htmlString, out c, out _);
```
MihaZupan, line 189: Could this throw if the entity represents an incomplete surrogate pair?

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -244,6 +325,64 @@ public readonly char PeekCharExtra(int offset)
...
 
+    /// <summary>
+    /// Peeks a rune at the specified offset from the current beginning of the slice
+    /// without using the range <see cref="Start"/> or <see cref="End"/>, returns default if outside the <see cref="Text"/>.
+    /// Recognizes supplementary code points that cannot be covered by a single character.
+    /// </summary>
+    /// <param name="offset">The offset.</param>
+    /// <returns>The rune at the specified offset, returns default if none.</returns>
+    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    internal readonly Rune PeekRuneExtra(int offset)
```
MihaZupan: I find the behavior of sometimes looking further than the specified offset odd for this generic helper.

Since this is only used to look at the previous character, can we change this to something like `Rune PreviousRune()` instead?

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -166,6 +211,42 @@ public char NextChar()
...
+            Start = End + 1;
+            return default;
+        }
+        var currentBmpOrHighSurrogate = Text[start++];
+        if (char.IsHighSurrogate(currentBmpOrHighSurrogate))
+        {
+            var currentLowSurrogate = Text[start++];
+            if (!char.IsLowSurrogate(currentLowSurrogate))
+            {
+                start--;
```
MihaZupan: Is there any test that's covering this?
Needing to backtrack in `NextRune` is odd and feels like a usage error, not something this method should be hiding.

MihaZupan:
> If we adopt C# 14, we can use [Extension Members](https://learn.microsoft.com/dotnet/csharp/whats-new/csharp-14#extension-members) instead of creating the UnicodeUtilitySupplement class.

Could you get by with making the existing `UnicodeUtlity` file `partial` instead?

src/Markdig/Helpers/CharHelper.cs
tats-u: Is it accessible by external plugins/extensions?
How about making the current `Rune` variant public only in .NET (where `Rune` publicly exists) by `#if NETCOREAPP3_0_OR_GREATER`, too?

```cs
#if NETCOREAPP3_0_OR_GREATER
public
#else
internal
#endif
static void CheckOpenCloseDelimiter(/* ... */)
```

It is also applied to all `Rune` variant APIs.

src/Markdig/Helpers/CharHelper.cs
tats-u: I suggest s/character to/code unit to/

In the spec, character is code _point_, more similar to `Rune`.

tats-u: Or "The character (code unit) to evaluate."?

src/Markdig/Helpers/CharHelper.cs
tats-u: Or should we make new APIs with `Rune` replaced with `uint` public for all modern and legacy .NET implementations? By this plugins/extensions will be able to bring their own `Rune` implementations through the `.Value` property.

tats-u:
> Could you get by with making the existing UnicodeUtlity file partial instead?

I moved the new members to the existing `UnicodeUtility.cs`. Should `UnicodeUtility.cs` also exist in the `Polyfills` directory with `partial` added to both class definitions?

src/Markdig/Helpers/CharHelper.cs
tats-u: i.e.

(1)

```suggestion
    /// <param name="c">The code unit to evaluate.</param>
```

or

(2)

```suggestion
    /// <param name="c">The character (code unit) to evaluate.</param>
```

or maybe

(3)

```suggestion
    /// <param name="c">The character (UTF-16 code unit) to evaluate.</param>
```

src/Markdig/Helpers/CharHelper.cs
tats-u:
```cs
    public static void CheckOpenCloseDelimiter(char pc, char c, bool enableWithinWord, out bool canOpen, out bool canClose)
    {
        /* ... */
    }

    public static void CheckOpenCloseDelimiter(uint pc, uint c, bool enableWithinWord, out bool canOpen, out bool canClose) => CheckOpenCloseDelimiter((Rune)pc, (Rune)c, enableWithinWord, out canOpen, out canClose);

#if NETCOREAPP3_0_OR_GREATER
    public
#else
    internal
#endif
    static void CheckOpenCloseDelimiter(Rune pc, Rune c, bool enableWithinWord, out bool canOpen, out bool canClose)
    {
        /* ... */
    }
```

MihaZupan: Making the Rune overload here conditionally public is fine.
I would avoid making the API worse (int) just for old targets.

src/Markdig/Helpers/CharHelper.cs
MihaZupan: IMO just "character" is clear enough since that's what they're usually called.

MihaZupan:
> Should UnicodeUtility.cs also exist in the Polyfills directory with partial added to both class definitions?

I don't think it's needed

src/Markdig/Helpers/CharHelper.cs
tats-u: I see. I'll this supplementary-character-friendly API dedicate to modern .NET targets for external plugins/extensions and to internal code.

src/Markdig/Helpers/CharHelper.cs
tats-u: If so I think we may need to tell what `c` in `Rune` and `uint` API means from this "character" notation. Are you going to make users guess what these "character"s actually means (UTF-16 code unit or scalar value) just/only from their types (`char` or `Rune`) (=unify the descriptions in the comments to "character")?

tats-u:
I see. Let me know if we should move it from `Helpers` to the `Polyfills` directory.

src/Markdig/Helpers/CharHelper.cs
tats-u: I'll add "A supplementary character is accepted" to Rune variants.

src/Markdig/Helpers/StringSlice.cs
tats-u: > find the behavior of sometimes looking further than the specified offset odd

I think it's a natural behavior if you replace the offending supplementary character (the valid surrogate pair) with a BMP character. We don't know which of a BMP or supplementary character the character at the given position is.

```
 -2  -1       0        1   2
[𠮷||  ]<we are here>[𠮷||  ]
    [吉]<we are here>[吉]

-----------------------------

High surrogate code unit
 ↓
[𠮷||  ]
     ↑
Low surrogate code unit
```

> [!NOTE]
> 𠮷 (U+20BB7) is a supplementary character and 吉 (U+5409) is a BMP character. 

What we need is just a character placed at the position. If the character occupies 2 UTF-16 code units, we should fetch the remaining outside half.

> Since this is only used to look at the previous character,

CJK Friendly Emphasis (#890) occasionally requires the previous character of the previous character of the current position. i.e. may need to call `PeekRuneExtra(-2)` or `PeekRuneExtra(-3)` in the future. It's not so bad idea to split that method into 2 (looking backward and forward).

tats-u: The reason why "Outdated" is displayed is simply because I changed the visibility of this method.

src/Markdig/Helpers/StringSlice.cs
tats-u: `String.Normalize` used by `TestParser.Compact` throws `ArgumentException` for strings containing isolated surrogate code units. I don't think we need to prepare one. Inputs whose outputs are accepted by `String.Normalize` don't go through this line.

tats-u:
@MihaZupan I have finished correcting all sections where your additional responses were not required.

src/Markdig/Parsers/Inlines/EmphasisInlineParser.cs
```diff
@@ -179,12 +180,16 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
...
+        Rune c = slice.CurrentRune;
 
         // The following character is actually an entity, we need to decode it
         if (HtmlEntityParser.TryParse(ref slice, out string? htmlString, out int htmlLength))
         {
-            c = htmlString[0];
+            var status = Rune.DecodeFromUtf16(htmlString, out c, out _);
+            if (status != System.Buffers.OperationStatus.Done)
+            {
+                throw new ArgumentException($"Could not extract the first Unicode scalar value (decode status: {status})", nameof(htmlString));
```
MihaZupan: My comment about throwing wasn't that we should be doing it manually, but that this shouldn't be throwing here.
If we hit an unpaired surrogate here, we can pick the behavior, e.g. skip matching in this parser, or treat the character as 0.

In general we shouldn't be throwing while parsing even if the input is weird.

src/Markdig/Helpers/StringSlice.cs
MihaZupan: Not all tests have to go through TestParser, we can test the StringSlice directly.
It's easy to make off by one errors here, or forget to move offsets correctly in edge cases, I think we should add the corresponding test coverage.

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -145,6 +174,33 @@ public readonly char this[int index]
...
+#if NETCOREAPP3_0_OR_GREATER
+    public
+#else
+    internal
+#endif
+    Rune RuneAt(int index)
+    {
+        var first = Text[index];
+        if (!char.IsSurrogate(first))
+            return new Rune(first);
```
MihaZupan: Rune constructors do their own validation.
Can we tweak these helpers to use the Create helpers instead?

E.g. something like this should have the same behavior.
```c#
if (!Rune.TryCreate(first, out Rune rune) && index + 1 <= End) 
{
    Rune.TryCreate(first, Text[index + 1], out rune);
}

return rune;
```

Same for other helpers

src/Markdig/Helpers/StringSlice.cs
MihaZupan: As an example, `RuneAt(int index)` is inconsistently validating the `End` right now. It's being checked for the second surrogate, but not the first character.

src/Markdig/Parsers/Inlines/EmphasisInlineParser.cs
tats-u: Do you know `Rune.DecodeFromUtf16` emits U+FFFD � if error?
I'll revert the last change though.

src/Markdig/Helpers/StringSlice.cs
tats-u: It should check `Text.Length` instead to align with the indexer. I added tests.

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -125,6 +126,33 @@ public readonly char CurrentChar
...
+            if (start > End) return default;
+            var first = Text[start];
+            // BMP character
+            if (Rune.TryCreate(first, out var rune)) return rune;
+            if (start + 1 > End) return default;
+            var second = Text[start + 1];
+            // Supplementary character
+            return Rune.TryCreate(first, second, out rune)
+                ? rune
+                : default;
```
MihaZupan:
```suggestion
            int start = Start;
            if (start > End) return default;

            char first = Text[start];
            if (!Rune.TryCreate(first, out Rune rune) && start + 1 <= End)
            {
                // The first character is a surrogate, check if we have a valid pair
                Rune.TryCreate(first, Text[start + 1], out rune);
            }

            return rune;
```

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -145,6 +173,34 @@ public readonly char this[int index]
...
+        if (Rune.TryCreate(first, out var rune))
+            return rune;
+        if (index + 1 < Text.Length)
+        {
+            var second = Text[index + 1];
+            return Rune.TryCreate(first, second, out rune)
+                ? rune
+                : default;
+        }
+        return default;
```
MihaZupan:
```suggestion
        string text = Text;
        char first = text[index];

        if (!Rune.TryCreate(first, out Rune rune) && (uint)(index + 1) < (uint)text.Length)
        {
            // The first character is a surrogate, check if we have a valid pair
            Rune.TryCreate(first, text[index + 1], out rune);
        }

        return rune;
```

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -166,6 +222,50 @@ public char NextChar()
...
+        if (Rune.TryCreate(first, out var rune))
+            return rune;
+        if (start + 1 > End)
+            return default;
+        var second = Text[start + 1];
+        // Supplementary character
+        return Rune.TryCreate(first, second, out rune)
+            ? rune
+            : default;
+    }
```
MihaZupan, line 266:
```suggestion
    Rune NextRune()
    {
        int start = Start;
        if (start >= End)
        {
            Start = End + 1;
            return default;
        }

        char first = Text[start++];
        if (!Rune.TryCreate(first, out Rune rune) && start <= End)
        {
            // The first character is a surrogate, check if we have a valid pair
            if (Rune.TryCreate(first, Text[start], out rune))
            {
                // Valid surrogate pair
                start++;
            }
        }
        
        Start = start;
        return rune;
    }
```

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -145,6 +173,34 @@ public readonly char this[int index]
...
+    /// </summary>
+    /// <param name="index">The index relative to the slice.</param>
+    /// <returns>The rune at the specified index or the default value (refers to <c>'\0'</c>) if the rune cannot be determined.</returns>    [MethodImpl(MethodImplOptions.AggressiveInlining)]
+    /// <exception cref="IndexOutOfRangeException">Thrown when the given <paramref name="index"/> is out of range</exception>
+#if NETCOREAPP3_0_OR_GREATER
+    public
+#else
+    internal
+#endif
+    Rune RuneAt(int index)
```
MihaZupan:
```suggestion
    readonly Rune RuneAt(int index)
```

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -244,6 +344,57 @@ public readonly char PeekCharExtra(int offset)
...
+        // The code unit at `index` should be a high surrogate and the start of a scalar value (rune) of a supplementary character
+        if (index + 1 >= text.Length)
+        {
+            return default;
+        }
+        var lowSurrogate = text[index + 1];
+        return Rune.TryCreate(bmpResultOrNearerSurrogate, lowSurrogate, out rune)
+            ? rune
+            : default;
+    }
```
MihaZupan, line 398:
```suggestion
    readonly Rune PeekRuneExtra(int offset)
    {
        int index = Start + offset;
        string text = Text;
        if ((uint)index >= (uint)text.Length)
        {
            return default;
        }

        char first = text[index];
        if (Rune.TryCreate(first, out var rune))
        {
            // BMP
            return rune;
        }

        // Check if we have a valid surrogate pair
        if (offset < 0)
        {
            // The code unit at `index` should be a low surrogate
            // The scalar value (rune) of a supplementary character should start at `index - 1`, which should be a high surrogate
            if ((uint)(index - 1) < (uint)text.Length)
            {
                Rune.TryCreate(text[index - 1], first, out rune);
            }
        }
        else
        {
            // The code unit at `index` should be a high surrogate and the start of a scalar value (rune) of a supplementary character
            if ((uint)(index + 1) < (uint)text.Length)
            {
                Rune.TryCreate(first, text[index + 1], out rune);
            }
        }

        return rune;
    }
```

src/Markdig.Tests/TestStringSlice.cs
```diff
@@ -0,0 +1,129 @@
...
+namespace Markdig.Tests;
+
+[TestFixture]
+internal class TestStringSlice
+{
+    // TODO: Add more tests for StringSlice
+
+    // Old versions of modern .NET (that meets NETCOREAPP but not the following #if) try to load the polyfilled Rune in .NET Standard 2.x build of Markdig while they have built-in Rune too
+    // Adjust this condition to match the minimal modern (not .NET Standard or Framework) .NET target framework of Markdig (not Markdig.Tests)
+#if NET8_0_OR_GREATER
```
MihaZupan: The tests have `InternalsVisibleTo`, so testing a netstandard build of Markdig on modern .NET doesn't work for these (see build errors from the GH actions).
Might have to use a `static using` to point it at the rune from `System.Runtime` (you can change the namespace of the polyfilled Rune).

src/Markdig/Helpers/StringSlice.cs
tats-u: > `if ((uint)(index - 1) < (uint)text.Length)`

`index` is never be 0 here (`IndexOutOfRangeException` has already been thrown at `text[index]`), but could you tell me why you prefer the cast into `uint` when compare indexes?

MihaZupan: The `uint` casts are a trick that allows the JIT to avoid inserting additional bounds checks for the length. It checks for "`index - 1` is not negative" and "`index - 1` is in range of the string Length" as a single comparison.

E.g. compare
https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAZgAJiAmcgFRlwwGEAbPXAWACgBvb8gZQrEAjEnIBLAHYZyAMQgQAFKIAM5XGkkztAExgIAlP0F8ugi5IBm5JdP0JycciPIAeDQDoAMjCkBzDAALQ0oAdg0AbXsDJxcAXQBuE0twl2TzQQBfFPJcskoxbVkAIWwoFRF1TWK9A2NMgTNUgQkbJSUAV2kMQzspBziRUI8unsNcHz9AkLTcaIHY5xEk3ItiCJEMixyuLKA=

tats-u: I see. Could you add an additional comment (e.g. `// index is never be 0 here (IndexOutOfRangeException would have already been thrown at text[index]) above`) above the `if` statement of the suggestion?

MihaZupan: Why can't it be 0? Wouldn't you hit it with `Start = 1`, `offset = -1`?

tats-u: I was under the wrong impression. `(uint)(index - 1)` will become ~4.3 billion (`uint.MaxValue`) when `index` is 0 due to underflow. It should be `(uint)index < (uint)text.Length + 1u` instead.

src/Markdig.Tests/TestStringSlice.cs
tats-u: We need to tweak the workflow file to add the following to .csproj:

```xml
<!-- *snip* -->
<PropertyGroup>
  <!-- *snip* -->
  <DefineConstants>$(DefineConstants);TEST_NETSTANDARD_IN_MODERN_NET</DefineConstants>
  <!-- *snip* -->
</PropertyGroup>
<!-- *snip* -->
```

Then, add the following to the test files that use `Rune`:

```cs
#if TEST_NETSTANDARD_IN_MODERN_NET
using Markdig.Polyfills.System.Text;
#else
using System.Text;
#endif
```

tats-u:
image

After accepting your suggestions, some tests now fail.

tats-u:
I managed to get all tests passed again.

src/Markdig/Helpers/StringSlice.cs
```diff
@@ -166,6 +225,50 @@ public char NextChar()
...
+        {
+            Start = End + 1;
+            return default;
+        }
+        // We don't need to Rune.TryCreate here because we don't use the created rune
+        if (
+            // Advance to the next character, checking for a valid surrogate pair
+            char.IsHighSurrogate(Text[start++])
+            // Don't unconditionally increment `start` here. Check the surrogate code unit at `start` is a part of a valid surrogate pair first.
+            && start <= End
```
tats-u, line 249: Is this covered by https://github.com/xoofx/markdig/pull/913/changes#diff-e66da161369de25c14420f6b83bc176a4b7240b03cee5f9478e67335faa6d0a7R243? (the above `if (start >= End)`)

tats-u:
I will need to merge `master` and fix the workflow.

tats-u:
I hope the new workflow works.

tats-u:
@MihaZupan I did all what I have to do now. Please let me know if I still have ones. The ball is in your court now. You can directly push commits here if necessary.

MihaZupan:
I opened a PR against your fork with some cleanup https://github.com/tats-u/markdig/pull/1

tats-u:
@MihaZupan Couldn't you directly push to my branch?

MihaZupan:
Sure, but for the most part I try not to do that to avoid breaking contributors' workflows by messing up their branch. Pressing merge on the PR in your repo has the same effect.

tats-u:
> Pressing merge on the PR in your repo has the same effect.

- "Create a merge commit" → an extra commit will be added
- "Squash and merge", "Rebase and merge" → the commit history is not compatible with your branch `MihaZupan:rune-feedback`

From now on, please feel free to add commits as needed, then just press “Squash and merge” at the end (all messy commits will be aggregated into a single commit).

tats-u:
Thank you for your approval.

xoofx:
> Thank you for your approval.

Thank for your work! I will do a release later tonight.

PR #916, opened 2025-11-28
Replace null checks with IsEmpty property for ReadOnlySpan<char>

prozolic:
This PR replace null checks with `IsEmpty` property for `ReadOnlySpan<char>`. 
The change suppresses [CA2265](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/quality-rules/ca2265) warnings.

### DIFF
--- a/src/Markdig/Helpers/LinkHelper.cs
+++ b/src/Markdig/Helpers/LinkHelper.cs
@@ -60,12 +60,12 @@ public static class LinkHelper
             }
             else
             {
-                normalized = allowOnlyAscii ? CharNormalizer.ConvertToAscii(c) : null;
+                normalized = allowOnlyAscii ? CharNormalizer.ConvertToAscii(c) : ReadOnlySpan<char>.Empty;
             }
 
             for (int j = 0; j < (normalized.Length < 1 ? 1 : normalized.Length); j++)
             {
-                if (normalized != null)
+                if (!normalized.IsEmpty)
                 {
                     c = normalized[j];
                 }


src/Markdig/Helpers/LinkHelper.cs
```diff
@@ -60,12 +60,12 @@ public static string Urilize(ReadOnlySpan<char> headingText, bool allowOnlyAscii
             }
             else
             {
-                normalized = allowOnlyAscii ? CharNormalizer.ConvertToAscii(c) : null;
+                normalized = allowOnlyAscii ? CharNormalizer.ConvertToAscii(c) : ReadOnlySpan<char>.Empty;
             }
 
             for (int j = 0; j < (normalized.Length < 1 ? 1 : normalized.Length); j++)
```
MihaZupan, line 66: Are we missing test coverage given that a null `normalized` would throw here?

prozolic: @MihaZupan Thank you for the review.
Since `normalized` is a `ReadOnlySpan<char>`, I changed the check from `normalized != null` to `!normalized.IsEmpty`.
When `CharNormalizer.ConvertToAscii(c)` returns `null`, it is implicitly converted to an empty span, maintaining the same behavior as before.
I confirmed from the following test cases.

https://github.com/xoofx/markdig/blob/master/src/Markdig.Tests/TestLinkHelper.cs#L330-L337

MihaZupan: Right, I was more referring to whether there are cases that could trigger an exception to be thrown before your current change. If so, it'd be good to have a test case for it.

prozolic: At this point, I cannot find any cases where an exception would be thrown before my change. 
If an empty string (`""`) were assigned to `normalized`, the results of `normalized != null` and `!normalized.IsEmpty` would differ, which could cause an exception. However, I cannot find any cases where `CharNormalizer.ConvertToAscii` would return an empty string (`""`).
Does this response address your concern?

PR #917, opened 2025-11-28
Use Dictionary.TryAdd instead of ContainsKey and indexer by reducing lookups.

prozolic:
This PR replace Dictionary.TryAdd instead of ContainsKey and indexer by reducing lookups.

### DIFF
--- a/src/Markdig/Parsers/ParserList.cs
+++ b/src/Markdig/Parsers/ParserList.cs
@@ -38,8 +38,10 @@ public abstract class ParserList<T, TState> : OrderedList<T> where T : notnull, 
             {
                 foreach (var openingChar in parser.OpeningCharacters)
                 {
-                    charCounter.TryAdd(openingChar, 0);
-                    charCounter[openingChar]++;
+                    if (!charCounter.TryAdd(openingChar, 1))
+                    {
+                        charCounter[openingChar]++;
+                    }
                 }
             }
             else


src/Markdig/Parsers/ParserList.cs
```diff
@@ -38,11 +38,10 @@ protected ParserList(IEnumerable<T> parsersArg) : base(parsersArg)
             {
                 foreach (var openingChar in parser.OpeningCharacters)
                 {
-                    if (!charCounter.ContainsKey(openingChar))
+                    if (!charCounter.TryAdd(openingChar, 1))
                     {
-                        charCounter[openingChar] = 0;
+                        charCounter[openingChar]++;
```
gfoidl, line 43: Here methods from `CollectionsMarshal` could be used (w/ a polyfill for the older targets), to avoid the double lookup (get, increment, set).

IBut, is it worth the churn here?

prozolic: @gfoidl  Thank you for the review!
As @gfoidl pointed out, using `CollectionsMarshal.GetValueRefOrAddDefault` can avoid double lookups.
I measured benchmarks before and after the changes, but it appears that significant performance improvements cannot be expected. Therefore, I think it may be better to hold off on adding this for now.


```csharp
            if (parser.OpeningCharacters is { Length: > 0 })
            {
                foreach (var openingChar in parser.OpeningCharacters)
                {
#if NET9_0_OR_GREATER
                    ref var countRef = ref CollectionsMarshal.GetValueRefOrAddDefault(charCounter, openingChar, out var exists);
                    if (exists)
                    {
                        countRef++;
                    }
                    else
                    {
                        countRef = 1; 
                    }
#else
                    if (!charCounter.TryAdd(openingChar, 1))
                    {
                        charCounter[openingChar]++;
                    }
#endif
                }
            }
```

<details>
<summary>Benchmark result</summary>

- PR

| Method                   | Mean     | Error     | StdDev    | Gen0   | Code Size | Gen1   | Allocated |
|------------------------- |---------:|----------:|----------:|-------:|----------:|-------:|----------:|
| PR | 2.472 μs | 0.0636 μs | 0.1876 μs | 0.8583 |  10,036 B | 0.0114 |   7.03 KB |
| CollectionsMarshal | 2.573 μs | 0.0988 μs | 0.2913 μs | 0.8583 |  10,808 B | 0.0114 |   7.03 KB |

</details>

<details>
<summary>Benchmark code</summary>
  
```csharp
[MemoryDiagnoser]
[DisassemblyDiagnoser(maxDepth: 3)]
public class ParserListBenchmark
{
    private List<BlockParser> _blockParsers;

    [GlobalSetup]
    public void Setup()
    {
        _blockParsers = new List<BlockParser>
        {
            new ThematicBreakParser(),
            new HeadingBlockParser(),
            new QuoteBlockParser(),
            new ListBlockParser(),
            new HtmlBlockParser(),
            new FencedCodeBlockParser(),
            new IndentedCodeBlockParser(),
            new ParagraphBlockParser(),
        };
    }

    [Benchmark(Description = "Create BlockParserList")]
    [MemoryRandomization]
    public BlockParserList CreateParserList()
    {
        return new BlockParserList(_blockParsers);
    }
}
```

</details>

MihaZupan: You'd see a perf difference if you hit the case where a character already existed in the dictionary. If you're just adding unique ones (as this benchmark seems to be), you're only doing one lookup in both cases.

I don't think it's worth the churn in this case, these aren't hot paths.

prozolic: @MihaZupan Thank you for the review!
Just to confirm, when you mentioned "these aren't hot paths", did you mean: the CollectionsMarshal optimization specifically, or the entire PR (TryAdd changes in all files)?
If it's the entire PR, I'll close it.

MihaZupan: I was mostly referring to `CollectionsMarshal` for this specific method.
If there were other usages, it does make things simpler and faster, e.g. your example above can be simplified to
```diff
-ref var countRef = ref CollectionsMarshal.GetValueRefOrAddDefault(charCounter, openingChar, out var exists);
-if (exists)
-{
-    countRef++;
-}
-else
-{
-    countRef = 1; 
-}
+CollectionsMarshal.GetValueRefOrAddDefault(charCounter, openingChar, out _)++;
```

I don't mind the `TryAdd` changes, but realistically only the one in `LinkReferenceDefinitionGroup` will get hit repeatedly.

prozolic: I see, then I'll keep only the original `TryAdd` changes and not add `CollectionsMarshal`.

> If there were other usages, it does make things simpler and faster, e.g. your example above can be simplified to

Thanks for showing me that pattern! I learned something new.

MihaZupan:
Thanks

PR #920, opened 2026-01-15
Support escaped quotes in generic attribute values

jnraine:
Hi! I've been using Markdig and it's been great. Wanted to contribute back with this small feature.

This is my first PR here so happy to make any revisions or tweaks as necessary. And also, if this isn't something wanted by the maintainers, feel free to close.

  ## Summary

  - Add backslash escape handling for quoted attribute values following kramdown's approach
  - This allows including quote characters within attribute values using backslash escapes (e.g. `data-value="hello \"world\""`)
  - It also works with single-quotes

  ## Motivation

  This was discovered when using markdown to generate Word documents from the Markdig AST. Word field codes use quoted strings:

  ```markdown
  [January 15, 2026](){field="DATE @ \"MMMM d, yyyy\""}
```

  This can work if the outer quotes are changed to single-quotes but escape sequences are useful for our context (and we preprocess the markdown so it works with Markdig today).

## Implementation

  Follows the same `hasEscape` pattern used in `LinkHelper.TryParseTitle`, using `IsAsciiPunctuation()` for escaping.

  This matches kramdown's behavior for inline attribute lists:

  > "If you need to use the value delimiter (a single or a double quote) inside the value, you need to escape it with a backslash."

  See: https://kramdown.gettalong.org/syntax.html

  ## Test plan

  - [x] Added 6 test cases covering escaped double/single quotes, escaped backslashes, and complex escape sequences

### DIFF
--- a/src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
+++ b/src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
@@ -203,49 +203,72 @@ public class GenericAttributesParser : InlineParser
                 int endValue = -1;
 
                 c = line.CurrentChar;
-                // Parse a quoted string
+                // Parse a quoted string with backslash escape support (following CommonMark conventions)
                 if (c == '\'' || c == '"')
                 {
+                    var valueBuilder = new ValueStringBuilder(stackalloc char[ValueStringBuilder.StackallocThreshold]);
                     char openingStringChar = c;
-                    startValue = line.Start + 1;
-                    while (true)
+                    bool hasEscape = false;
+                    while ((c = line.NextChar()) != '\0')
                     {
-                        c = line.NextChar();
-                        if (c == '\0')
+                        if (hasEscape)
                         {
-                            return false;
+                            hasEscape = false;
+                            // Following CommonMark: only ASCII punctuation can be escaped
+                            if (!c.IsAsciiPunctuation())
+                            {
+                                valueBuilder.Append('\\');
+                            }
+                            valueBuilder.Append(c);
                         }
-                        if (c == openingStringChar)
+                        else if (c == openingStringChar)
                         {
                             break;
                         }
+                        else if (c == '\\')
+                        {
+                            hasEscape = true;
+                        }
+                        else
+                        {
+                            valueBuilder.Append(c);
+                        }
                     }
-                    endValue = line.Start - 1;
-                    c = line.NextChar(); // Skip closing opening string char
+
+                    if (c == '\0')
+                    {
+                        valueBuilder.Dispose();
+                        return false;
+                    }
+
+                    properties ??= new();
+                    properties.Add(new KeyValuePair<string, string?>(name, valueBuilder.ToString()));
+                    valueBuilder.Dispose();
+
+                    line.SkipChar(); // Skip closing quote
+                    c = line.CurrentChar;
+                    continue;
                 }
-                else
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs [hunk 1]
```diff
@@ -203,49 +203,72 @@ public static bool TryParse(ref StringSlice slice, [NotNullWhen(true)] out HtmlA
...
+                        valueBuilder.Dispose();
+                        return false;
+                    }
+
+                    properties ??= new();
+                    properties.Add(new KeyValuePair<string, string?>(name, valueBuilder.ToString()));
+                    valueBuilder.Dispose();
+
+                    line.SkipChar(); // Skip closing quote
+                    c = line.CurrentChar;
```
MihaZupan, line 249:
```suggestion
                    c = line.NextChar(); // Skip closing quote
```

src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
```diff
@@ -203,49 +203,72 @@ public static bool TryParse(ref StringSlice slice, [NotNullWhen(true)] out HtmlA
                 int endValue = -1;
 
                 c = line.CurrentChar;
-                // Parse a quoted string
+                // Parse a quoted string with backslash escape support (following CommonMark conventions)
                 if (c == '\'' || c == '"')
                 {
+                    var valueBuilder = new ValueStringBuilder(stackalloc char[ValueStringBuilder.StackallocThreshold]);
```
MihaZupan, line 209: Since you can't stackalloc in a loop, this'll have to either be `new ValueStringBuilder([])`, or move the search logic into a separate helper method.

src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
```diff
@@ -203,49 +203,72 @@ public static bool TryParse(ref StringSlice slice, [NotNullWhen(true)] out HtmlA
...
+
+                    if (c == '\0')
+                    {
+                        valueBuilder.Dispose();
+                        return false;
+                    }
+
+                    properties ??= new();
+                    properties.Add(new KeyValuePair<string, string?>(name, valueBuilder.ToString()));
+                    valueBuilder.Dispose();
```
MihaZupan, line 246:
```suggestion
```
(the ToString already disposes)

src/Markdig/Extensions/GenericAttributes/GenericAttributesParser.cs
```diff
@@ -203,49 +203,72 @@ public static bool TryParse(ref StringSlice slice, [NotNullWhen(true)] out HtmlA
...
-                    if (!valid)
+
+                    if (c == '\0')
+                    {
+                        valueBuilder.Dispose();
+                        return false;
+                    }
+
+                    properties ??= new();
+                    properties.Add(new KeyValuePair<string, string?>(name, valueBuilder.ToString()));
```
MihaZupan, line 245: IMO it might be a bit easier to read if you change the structure to something like
```c#
string value;

if (c == '\'' || c == '"')
{
    // ...
    value = valueBuilder.ToString();
}
else
{
    int start = line.Start;
    // ...
    value = slice.Text.Substring(...);
}

properties ??= new();
properties.Add(new KeyValuePair<string, string?>(name, value));
```

PR #921, opened 2026-01-21
Add CJK-friendly Emphasis Extension

### LINKED ISSUE #890: Support CJK Friendly Emphasis extension

tats-u:
Closes #890


src/Markdig.Tests/Specs/CJKFriendlyEmphasis.md
```diff
@@ -0,0 +1,37 @@
+## CJK-friendly Emphasis Extension
```
MihaZupan, line 1: Please add this to the list of spec files here: https://github.com/tats-u/markdig/blob/e0aa9beabe185c59dac2d585dcce1d1677a5ff6d/src/SpecFileGen/Program.cs#L61
so that these get turned into generated test cases

Or alternatively just as a separate test file with these cases if you find that easier.

src/Markdig/MarkdownExtensions.cs
```diff
@@ -522,6 +523,12 @@ public static MarkdownPipelineBuilder UseGlobalization(this MarkdownPipelineBuil
         return pipeline;
     }
 
+    public static MarkdownPipelineBuilder UseCjkFriendlyEmphasis(this MarkdownPipelineBuilder pipeline)
+    {
+        pipeline.Extensions.AddIfNotAlready<CjkFriendlyEmphasisExtension>();
```
MihaZupan: I don't think this needs to go through an extension.
If all we need is to change an option on the emphasis parser, I'd make this helper just do that directly.

MihaZupan (commented):
Thanks, I'll try to take a closer look in a few days.

Do you have any feeling about whether the spec has willingness to adopt something like this out of the box?
Ideally something like this wouldn't even be an option, just the default behavior from the spec.

tats-u:
> Do you have any feeling about whether the spec has willingness to adopt something like this out of the box?
> Ideally something like this wouldn't even be an option, just the default behavior from the spec.

I think that "Out of the box" is a little bit too early for a while. It is not _strictly_ compatible with CommonMark. It is a provisional spec and probably going to be revised once more (to peek the character two ahead to deal with some following variation sequences)
It is fine to strongly recommend this.

src/Markdig/MarkdownExtensions.cs
tats-u: This is substantially a parser option as you suspect. I chose an extension because it is natural to be treated as for users.

MihaZupan: I think it's okay for it to be an _extension method_ on the pipeline builder, but I don't see the value in going through an `IMarkdownExtension`. We could always add it later if there is a need.

src/Markdig/Parsers/Inlines/EmphasisInlineParser.cs
```diff
@@ -189,8 +203,17 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             Rune.DecodeFromUtf16(htmlString, out c, out _);
         }
 
+        bool canOpen = false;
+        bool canClose = false;
```
MihaZupan:
```suggestion
        bool canOpen;
        bool canClose;
```

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -150,6 +150,113 @@ private static void CheckOpenCloseDelimiter(bool prevIsWhiteSpace, bool prevIsPu
...
+        bool eitherIsCjk = prevIsCjk || nextIsCjk;
+
+        canOpen |= eitherIsCjk || !nextIsPunctuation;
+        canClose |= eitherIsCjk || !prevIsPunctuation;
+
+        static bool IsNonEmojiGeneralUseVariantSelector(Rune r) => r.Value is >= 0xFE00 and <= 0xFE0E;
+        static bool IsIdeographicVariationSelector(Rune r) => r.Value is >= 0xE0100 and <= 0xE01EF;
+        static bool IsCjkAmbiousPunctuation(Rune main, Rune vs) => vs.Value is 0xFE01 && main.Value is 0x2018 or 0x2019 or 0x201C or 0x201D;
+        // As of Unicode 17
+        static bool IsCjk(Rune r) => r.Value is
```
MihaZupan, line 214: Please add a link to where these constants are coming from

src/Markdig/Helpers/CharHelper.cs
```diff
@@ -150,6 +150,113 @@ private static void CheckOpenCloseDelimiter(bool prevIsWhiteSpace, bool prevIsPu
         }
     }
 
+#if NET
+    public
+#else
+    internal
+#endif
+    static void CheckOpenCloseDelimiterCjkFriendly(Rune pc, Rune c, Rune twoPreviousRune, bool enableWithinWord, out bool canOpen, out bool canClose)
```
MihaZupan: We probably need some target tests for this method.
I won't pretend to understand all the CJK implications here, but it's not immediately obvious to me e.g. why it's okay to skip further checks if we have a space before/after.

tats-u: > e.g. why it's okay to skip further checks if we have a space before/after.

I'll add a link to https://github.com/tats-u/markdown-cjk-friendly/commit/3c4217bea8248e9abc8be4e7c68748a88557662d

src/Markdig/Helpers/CharHelper.cs
tats-u: Do you mean:

- https://github.com/tats-u/markdown-cjk-friendly/blob/main/ranges.md
- https://github.com/tats-u/markdown-cjk-friendly/blob/main/specification.md

src/Markdig/MarkdownExtensions.cs
tats-u: I removed that class. (kept that extension method)

tats-u:
How can we refer to the polyfilled `Rune` easily?

```xml
    <ProjectReference Include="..\Markdig\Markdig.csproj">
      <Aliases>global</Aliases>
    </ProjectReference>
```

forces us to add extra `Markdig::` everywhere.

MihaZupan:
I'd ifdef the test to just NET

tats-u:
I have no where should be `#ifdef NET`'ed.

MihaZupan:
Does something like this not work?
```c#
#if NET
[Test]
[TestCase(...)]
public void TestCheckOpenCloseDelimiterCjkFriendly(...)
{
    // ...
}
#endif
```

tats-u:
As I suspected, it is not concerned with the target framework of the main project. It only refers to the target of the testing project.

{C8CB7325-9E0B-49DE-B99A-06038A93D0F0}

tats-u:
Just resolved the conflict.

@MihaZupan If we try to minimize the amount of changes, I think that we cannot help adding `uint` APIs (not only the current `Rune` ones).

xoofx:
> @MihaZupan If we try to minimize the amount of changes, I think that we cannot help adding `uint` APIs (not only the current `Rune` ones).

It might be acceptable to make this feature only possible for `>= net8.0` instead.

MihaZupan:
@tats-u I'd prefer we avoid exposing APIs specific to old TFMs.
I'll try to check over next weekend how we could get the tests working here.

tats-u:
In that case, is it fine to completely exclude tests directly using Rune in the netstandard2.x tests in the CI?

src/Markdig/Parsers/Inlines/EmphasisInlineParser.cs
```diff
@@ -43,6 +43,12 @@ public EmphasisInlineParser()
     /// </summary>
     public List<EmphasisDescriptor> EmphasisDescriptors { get; }
 
+    /// <summary>
+    /// Gets or toggles whether the emphasis parser should be CJK-friendly.
+    /// </summary>
+    /// <seealso href="https://github.com/tats-u/markdown-it-cjk"/>
```
lindexi: `https://github.com/tats-u/markdown-it-cjk` is 404

Image

tats-u: Oh, no. Fixed

tats-u:
I'd like give up testing methods that use Rune in .NET Standard 2.x (lib) + .NET (test)

xoofx:
> In that case, is it fine to completely exclude tests directly using Rune in the netstandard2.x tests in the CI?

Yes for now it is ok, so that we will be able to merge your request.

tats-u:
I'll give it a try for a while, but I'll revert it if it doesn't work out.

tats-u:
I will discard .NET Framework adoption commits.

src/Markdig/MarkdownExtensions.cs
```diff
@@ -515,6 +515,18 @@ public static MarkdownPipelineBuilder UseGlobalization(this MarkdownPipelineBuil
         return pipeline;
     }
 
+/// <summary>
```
xoofx: Leading 4 spaces missing?

tats-u: I wonder why I forgot it. Thank you

xoofx:
Thank you @tats-u for the hard work!

tats-u:
Thank you for your adoption.

tats-u:
I will merge https://github.com/tats-u/markdown-cjk-friendly/pull/32 to introduce Mardig in the README of the extension after you release a new version of Markdig.

xoofx:
> I will merge [tats-u/markdown-cjk-friendly#32](https://github.com/tats-u/markdown-cjk-friendly/pull/32) to introduce Mardig in the README of the extension after you release a new version of Markdig.

Markdig 1.1 is released and your work has a dedicated site page: https://xoofx.github.io/markdig/docs/extensions/cjk-friendly-emphasis/

tats-u:
Wow, thank you

PR #923, opened 2026-02-09
Add spoiler extension

### LINKED ISSUE #596: Support for Spoiler
Is there any support for spoiler?

Maybe like StackOverflow spoiler using `>!` which is basically just a `<blockquote>` with a `.spoiler` class?

Abrynos:
Adds inline spoilers to fix #596 (last comment as of today, which is where inlines were requested).

Need ideas from the CSS-pros here on how to nicely format this when the **bootstrap** extension is active.

The effects of following .css would surely look nice, but I'm not 100% sure how to achieve it using bootstrap and with as little code as possible.

```css
.spoiler {
  color: black;
  background-color: black;
}
.spoiler:hover {
  background-color: white;
}
```

### DIFF
--- a/src/Markdig/Extensions/Spoilers/SpoilerExtension.cs
+++ b/src/Markdig/Extensions/Spoilers/SpoilerExtension.cs
@@ -1,0 +1,36 @@
+// Copyright (c) Alexandre Mutel. All rights reserved.
+// This file is licensed under the BSD-Clause 2 license.
+// See the license.txt file in the project root for more information.
+
+using Markdig.Parsers.Inlines;
+using Markdig.Renderers;
+using Markdig.Renderers.Html;
+
+namespace Markdig.Extensions.Spoilers;
+
+/// <summary>
+///     Extension for parsing inline spoilers starting and ending with an equal amount of pipe characters.
+/// </summary>
+/// <seealso cref="IMarkdownExtension"/>
+public sealed class SpoilerExtension : IMarkdownExtension
+{
+    /// <inheritdoc />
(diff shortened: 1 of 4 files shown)


Abrynos:
Something like this could work, utilizing bootstrap variables, but it's very explicit:
```html
<span class="spoiler" style="{ color: var(--bs-gray); background-color: var(--bs-gray); display: inline-block; border-radius: 15%; } :hover { color: var(--bs-primary); background-color: transparent; }">content</span>
```
Any thoughts? @xoofx @Atulin

Atulin:
Not a huge fan of depending on Bootstrap tbh. I think I'd just emit `<span class="md-spoiler">` or some such and leave the styling to the user. Maybe leave it in the plugin options

xoofx:
Thank you @Abrynos but I'm going to decline this. The rationale is:

- The original syntax in Markdig (::: and :: for div, was inspired by [Pandoc fenced divs](https://pandoc.org/MANUAL.html#extension-fenced_divs). In general, I don't want to integrate a syntax that is not supported by another well-known parser (this one though was niche, I agree)
- GitHub does not support the syntax proposed here but they support https://gist.github.com/jbsulli/03df3cdce94ee97937ebda0ffef28287 which is completely fine also to use in the markdown files and the recommended syntax, as they will display correctly on GitHub, and you can easily display them correctly on your website with the appropriate css.

Atulin:
Neither of those are inline, but yeah, I can understand leaving non-standard syntax for plugins to implement

Issue #807, opened 2024-07-12
Emoji in table


drshihan:
In a table using markdown, emojis are not rendered but their command text is displayed.

```
:x: // this works

// this doesn't work
|test|
|-|
|:x:|
```

prozolic:
This issue has been Fixed by #928.

PR #928, opened 2026-03-04
Fix emoji not rendering in table cells

prozolic:
This PR added `char.IsLetterOrDigit` to the character check in `EmojiParser.Match`.
It enables rendering for emoji patterns like `|:x:|`.
#807


xoofx:
Good catch! Indeed, this is something that should be fixed, thank you!

Issue #846, opened 2025-01-03
Url and title spans are set incorrectly for LinkReferenceDefinition inside a quote block.


snnz:
For example, for the following markdown:
```
> [foo]:
> /url 'the title'
>
> [foo]
```

Roundtrip renderer, that relies on those spans, produces this:
```
> [foo]:
> /ur' 'the tit'
> 
>[foo]
```

prozolic:
This issue has been Fixed by #931

PR #931, opened 2026-03-14
Fix to calculate LinkReferenceDefinition span positions from StringLineGroup.Lines

prozolic:
This PR fixes to calculate `LinkReferenceDefinition` span positions from `StringLineGroup.Lines` instead of `lines.Lines[0].Slice.Start`.
Also fix `TextRendererBase.Write(ReadOnlySpan<char>)` to update `previousWasLine` when the content ends with newline.
#846

### DIFF
--- a/src/Markdig/Parsers/ParagraphBlockParser.cs
+++ b/src/Markdig/Parsers/ParagraphBlockParser.cs
@@ -212,13 +212,11 @@ public class ParagraphBlockParser : BlockParser
 
                 // Correct the locations of each field
                 linkReferenceDefinition.Line = lines.Lines[0].Line;
-                int startPosition = lines.Lines[0].Slice.Start;
 
-                linkReferenceDefinition.Span        = linkReferenceDefinition.Span      .MoveForward(startPosition);
-                linkReferenceDefinition.LabelSpan   = linkReferenceDefinition.LabelSpan .MoveForward(startPosition);
-                linkReferenceDefinition.UrlSpan     = linkReferenceDefinition.UrlSpan   .MoveForward(startPosition);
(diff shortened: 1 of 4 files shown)


src/Markdig/Parsers/ParagraphBlockParser.cs
```diff
@@ -292,4 +289,31 @@ private static bool TryMatchLinkReferenceDefinitionTrivia(ref StringLineGroup li
 
         return atLeastOneFound;
     }
+
+    private static SourceSpan ConvertToAbsoluteSpan(StringLineGroup lines, SourceSpan span)
```
xoofx: Might be better to add it to `StringLineGroup`. We can keep them internal for now.

prozolic: Agreed. I've moved them as internal methods to `StringLineGroup`.

src/Markdig/Renderers/TextRendererBase.cs
```diff
@@ -320,6 +320,10 @@ public void Write(ReadOnlySpan<char> content)
         {
             WriteIndent();
             WriteRaw(content);
+            if (content[content.Length - 1] == '\n' || content[content.Length - 1] == '\r')
```
MihaZupan: Do we have any tests where the `\r` matters? It seems odd this would differ from `Write(char)`.

prozolic: You're right, `Write(char)` only checks for '\n'. I added '\r' in an intermediate fix, but that was incorrect .
I've fixed it.

PR #936, opened 2026-04-18
Fix AbbreviationExtension corrupting emphasis/bold/italic resolution (#935)

### LINKED ISSUE #935: AbbreviationExtension corrupts emphasis (bold/italic) resolution on the same line
## Description

When an abbreviation is defined and appears in the same paragraph as `**bold**` or `*italic*` markers, the emphasis is not resolved. The `**`/`*` characters are rendered as literal text instead.

## Minimal Reproducible Example

```csharp
#:package Markdig@1.1.*

using Markdig;
using Markdig.Syntax;
using Markdig.Syntax.Inlines;
using Markdig.Extensions.Abbreviations;

var md = """
HTML supports **bold** and *italic*

*[HTML]: HyperText Markup Language
""";

var pipeline = new MarkdownPipelineBuilder().UseAbbreviations().Build();
var doc = Markdown.Parse(md, pipeline);

void DumpInline(Inline inline, int depth = 0)
{
    var indent = new string(' ', depth * 2);
    switch (inline)
    {
        case LiteralInline lit:       Console.WriteLine($"{indent}LiteralInline: \"{lit.Content}\""); break;
        case AbbreviationInline abbr: Console.WriteLine($"{indent}AbbreviationInline: \"{abbr.Abbreviation?.Label}\""); break;
        case EmphasisInline em:
            Console.WriteLine($"{indent}EmphasisInline (char='{em.DelimiterChar}' count={em.DelimiterCount}):");
            foreach (var child in em) DumpInline(child, depth + 1);
            break;
        default:
            Console.WriteLine($"{indent}{inline.GetType().Name}");
            if (inline is ContainerInline ci)
                foreach (var child in ci) DumpInline(child, depth + 1);
            break;
    }
}

foreach (var block in doc)
{
    Console.WriteLine($"[{block.GetType().Name}]");
    if (block is LeafBlock lb && lb.Inline != null)
        foreach (var inline in lb.Inline) DumpInline(inline);
}
```

## Actual AST (with `UseAbbreviations()`)

```
[ParagraphBlock]
ContainerInline
  AbbreviationInline: "HTML"
  LiteralInline: " supports "
  EmphasisDelimiterInline
    LiteralInline: "bold"
    EmphasisDelimiterInline
      LiteralInline: " and "
      EmphasisDelimiterInline
        LiteralInline: "italic"
        EmphasisDelimiterInline
```

All `EmphasisDelimiterInline` nodes remain unresolved — no `EmphasisInline` is ever created.

## Expected AST (without `UseAbbreviations()`)

```
[ParagraphBlock]
LiteralInline: "HTML supports "
EmphasisInline (char='*' count=2):
  LiteralInline: "bold"
LiteralInline: " and "
EmphasisInline (char='*' count=1):
  LiteralInline: "italic"
```

## Root Cause Hypothesis

The `AbbreviationParser` hooks `LiteralInlineParser.PostMatch` and, when a match is found, replaces `processor.Inline` with a newly-created `ContainerInline`. When this fires on the first literal of a paragraph (e.g. `"HTML supports "`), the `ContainerInline` becomes the current parent context for subsequent inline parsing. The `EmphasisDelimiterInline` nodes that follow are then appended as children of that `ContainerInline` rather than as top-level siblings of the paragraph's inline chain. The emphasis post-processor, which resolves delimiter pairs within a shared parent, can no longer match the opening and closing `**`/`*` delimiters, so they remain unresolved.

The problem only manifests when the abbreviation appears **before** emphasis markers on the same line (i.e. the PostMatch fires and returns a `ContainerInline` before the emphasis delimiters have been parsed).

## Environment

- Markdig 1.1.x
- .NET 10

Kryptos-FR:
Closes #935

## Summary

`AbbreviationParser` subscribed to `LiteralInlineParser.PostMatch` to perform abbreviation substitution *during* inline parsing, before emphasis delimiters were resolved. This caused two bugs:

- Abbreviations before emphasis on the same line left `**`/`*` as literal text instead of resolving them.
- Abbreviations inside emphasis (e.g. `**HTML is great**`) were silently dropped due to an off-by-one in the word-boundary guard (`i != 0` instead of `i != content.Start`).

The fix moves substitution to `document.ProcessInlinesEnd`, so it runs after the full inline tree — including emphasis and links — has been resolved. Abbreviations are inserted as direct siblings into each parent container rather than wrapped in an intermediate `ContainerInline`.

## Testing

Three new spec examples cover the two bugs and their combination. All 3 764 existing tests pass.

## Performance

Benchmarks run on .NET 10 / Intel Core i3-10110U. "Before" = `upstream/main` (old `PostMatch` approach); "after" = this PR.

| Scenario | Before (mean) | After (mean) | Δ time | Before alloc | After alloc | Δ alloc |
|---|---:|---:|---:|---:|---:|---:|
| 2 defs, 10 paragraphs | 9.743 µs | 8.260 µs | **−15 %** | 8.19 KB | 8.14 KB | −1 % |
| 2 defs, 200 paragraphs | 157.1 µs | 141.4 µs | **−10 %** | 124.0 KB | 119.6 KB | −4 % |
| 10 defs, 10 paragraphs | 14.41 µs | 13.41 µs | **−7 %** | 16.81 KB | 16.34 KB | −3 % |
| 10 defs, 200 paragraphs | 243.9 µs | 222.8 µs | **−9 %** | 247.7 KB | 234.7 KB | **−5 %** |
| Defs present, no matches, 200 paragraphs | 126.8 µs | 125.1 µs | −1 % (noise) | 78.7 KB | 82.9 KB | +5 % |

Key gains come from eliminating `processor.GetSourcePosition()` (a binary search per match, now replaced by O(1) arithmetic) and removing the `ContainerInline` wrapper allocation per matched literal. The slight allocation uptick in the no-matches case is from the `Stack<Block>` used by `document.Descendants<LeafBlock>()`, which fires even when nothing is substituted; this is within noise for real documents.

### DIFF
--- a/src/Markdig.Tests/Specs/AbbreviationSpecs.generated.cs
+++ b/src/Markdig.Tests/Specs/AbbreviationSpecs.generated.cs
@@ -220,5 +220,59 @@ namespace Markdig.Tests.Specs.Abbreviations
 
             TestParser.TestSpec("*[Foo]: foo\n*[Foo Bar]: foobar\n\nFoo B", "<p><abbr title=\"foo\">Foo</abbr> B</p>", "abbreviations|advanced", context: "Example 11\nSection Extensions / Abbreviation\n");
         }
+
+        // Abbreviation before emphasis on the same line should be resolved:
+        [Test]
+        public void ExtensionsAbbreviation_Example012()
+        {
+            // Example 12
+            // Section: Extensions / Abbreviation
+            //
+            // The following Markdown:
+            //     *[HTML]: HyperText Markup Language
+            //     
+            //     HTML supports **bold** and *italic*
+            //
+            // Should be rendered as:
+            //     <p><abbr title="HyperText Markup Language">HTML</abbr> supports <strong>bold</strong> and <em>italic</em></p>
+
+            TestParser.TestSpec("*[HTML]: HyperText Markup Language\n\nHTML supports **bold** and *italic*", "<p><abbr title=\"HyperText Markup Language\">HTML</abbr> supports <strong>bold</strong> and <em>italic</em></p>", "abbreviations|advanced", context: "Example 12\nSection Extensions / Abbreviation\n");
+        }
+
+        // Abbreviation inside emphasis should be resolved:
+        [Test]
+        public void ExtensionsAbbreviation_Example013()
+        {
+            // Example 13
+            // Section: Extensions / Abbreviation
+            //
+            // The following Markdown:
+            //     *[HTML]: HyperText Markup Language
+            //     
+            //     **HTML is great**
+            //
+            // Should be rendered as:
+            //     <p><strong><abbr title="HyperText Markup Language">HTML</abbr> is great</strong></p>
+
+            TestParser.TestSpec("*[HTML]: HyperText Markup Language\n\n**HTML is great**", "<p><strong><abbr title=\"HyperText Markup Language\">HTML</abbr> is great</strong></p>", "abbreviations|advanced", context: "Example 13\nSection Extensions / Abbreviation\n");
+        }
+
+        // Abbreviation both before and inside emphasis on the same line should be resolved:
+        [Test]
+        public void ExtensionsAbbreviation_Example014()
+        {
+            // Example 14
+            // Section: Extensions / Abbreviation
+            //
+            // The following Markdown:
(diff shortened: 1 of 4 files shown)


xoofx:
Thanks for the fix!

Issue #946, opened 2026-08-25
A paragraph ending with `):` that should be in bold is not rendered correctly


bstordrup:
I have a setup where I add emojis and smilys (using `UseEmojiAndSmiley()`), but that sometimes collide with regular inline emphasis (especially bold: `**`).

For example this one:

```
**Non-goals (explicitly out of scope for this plan):**
```

is being rendered to 

<img width="429" height="44" alt="Image" src="https://github.com/user-attachments/assets/75c715a7-4d52-4fa6-8059-9799afddda2b" />

As I see it, this behavior occur because the `EmojiParser` is being executed before the `EmphasisInlineParser`, and then the `):*` is treated as an emoji. If the `EmphasisInlineParser` was executed first, then the whole line would become bold, and the `):*` would not be picked up as an emoji.

A workaround would be to ensure that there was a space before the closing `**`, but if I do not control the source markdown (my scenario is in a viewer app - not an editor), this may not be possible.

bstordrup:
Hmmm. Handling InlineParsers first does not solve it, but might be that the `EmojiParser` should instead look at the characters **after** the found Emoji Short Code if the last character in the Emoji Short Code is a `*`. If the next characters do **not** contain a full `EmphasisDescriptor('*', 1, 2, true)`, the emoji should be considered a false positve and skipped by the parser.

bstordrup:
Debugging into the parsers, and the problem is that the `EmojiParser.Match` is invoked with a text containing `:**`.

This will match the registered short code emoji `:*`, but it should not, because that breaks the `**` Emphasis for terminating a bold rendering.

Attaching a short markdown illustrating the issue.

[TestMarkdown.md](https://github.com/user-attachments/files/31423288/TestMarkdown.md)

PR #947, opened 2026-08-25
Reject emoji-parsing if the slice ends with `**`

bstordrup:
This pull request introduces an additional validation step in the emoji parsing logic to prevent false positives. Specifically, it ensures that emoji candidates ending with two asterisks (`**`) are not matched as valid emojis.

Emoji parsing improvement:

* Updated the `Match` method in `EmojiParser.cs` to return `false` if an emoji candidate ends with two asterisks, preventing incorrect emoji matches in such cases.

A sample markdown is provided on the #946 issue.

Closing #946

### DIFF
--- a/src/Markdig/Extensions/Emoji/EmojiParser.cs
+++ b/src/Markdig/Extensions/Emoji/EmojiParser.cs
@@ -36,17 +36,26 @@ public class EmojiParser : InlineParser
             return false;
         }
 
-        // An emoji candidate must not end with two '*'.
-        var lastChar = slice.PeekChar(slice.Length - 1);
-        var secondLastChar = slice.PeekChar(slice.Length - 2);
-        if (lastChar == '*' && secondLastChar == '*')
-        {
-            return false;
-        }
+        //// An emoji candidate must not end with two '*'.
+        //var lastChar = slice.PeekChar(slice.Length - 1);
+        //var secondLastChar = slice.PeekChar(slice.Length - 2);
+        //if (lastChar == '*' && secondLastChar == '*')
+        //{
+        //    return false;
+        //}
 
         // Try to match an emoji shortcode or smiley
         if (!_emojiMapping.PrefixTree.TryMatchLongest(slice.Text.AsSpan(slice.Start, slice.Length), out KeyValuePair<string, string> match))
         {
+            return false;
+        }
+
+        // An emoji candidate must not end with two '*'.
+        var lastChar = match.Key[match.Key.Length - 1]; // slice.PeekChar(slice.Length - 1);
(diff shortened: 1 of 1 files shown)


src/Markdig/Extensions/Emoji/EmojiParser.cs
```diff
@@ -42,6 +42,19 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             return false;
         }
 
+        // If the found emoji ends with a ´*´, the following char must not be a '*'.
+        var lastEmojiChar = match.Key[match.Key.Length - 1];
+        if (lastEmojiChar == '*')
+        {
+            // Only look at the following char if the emoji ends with a '*', otherwise it is not needed.
+            var followingChar = slice.PeekChar(match.Key.Length);
+            if (lastEmojiChar == '*' && followingChar == '*')
```
MihaZupan: `lastEmojiChar == '*'` is already checked above

src/Markdig/Extensions/Emoji/EmojiParser.cs
```diff
@@ -42,6 +42,19 @@ public override bool Match(InlineProcessor processor, ref StringSlice slice)
             return false;
         }
 
+        // If the found emoji ends with a ´*´, the following char must not be a '*'.
```
MihaZupan, line 45: Your current example is
`**Non-goals (explicitly out of scope for this plan):**`
but why would we treat that differently for emoji compared to italic
`*Non-goals (explicitly out of scope for this plan):*`
or other types of emphasis?

MihaZupan (commented):
Please add tests for any such behavior changes

src/Markdig/Extensions/Emoji/EmojiParser.cs
bstordrup: Which parser detects italic?

MihaZupan: https://github.com/xoofx/markdig/blob/main/src/Markdig/Parsers/Inlines/EmphasisInlineParser.cs handles this, and it's costumizable.
E.g. see https://github.com/xoofx/markdig/blob/main/src/Markdig/Extensions/EmphasisExtras/EmphasisExtraExtension.cs

bstordrup: Hmmm. I see the point. The EmphasisDescriptor takes a minumum and maximum count - bold is maximum count and italic is minimum count.

So actually the EmojiParser need to know if an EmphasisParser has open emphasis or not - and reject the emoji ending in ´*´ if there is an open emphasis at the point of it.

Issue #482, opened 2020-10-04
Bug: NormalizeRenderer incorrectly renders `- > p`


generateui:
`- > p` renders `- p`, where I expect the NormalizeRenderer to render `- > p`

xoofx:
Yes, there are many stuffs not working in the `NormalizeRenderer`. There was an opened issue #155 and PR #184  to improve the support, but it has never been finished. No time to fix it or work on it.

boxofyellow:
I was looking for issue to try and fix, but I was not able to reproduce this.

`Issues482.csproj`
```xml
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net10.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="Markdig" Version="0.3.0" />
  </ItemGroup>

</Project>
```

`Program.cs`
```C#
using Markdig;

var text = @"- > p";

var pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().Build();
Console.WriteLine(Markdown.ToHtml(text, pipeline));
```

Yields

```html
<ul>
<li>
<blockquote class="blockquote">
<p>p</p>
</blockquote>
</li>
</ul>
```

I tried it all the way back to 0.3.0 (prior to that it would fail to compile)

MihaZupan:
You're looking at `ToHtml`, not the normalize logic this issue is describing

boxofyellow:
Thanks for calling that out, I'll try again.

boxofyellow:
Yeah this seems to reproduce the problem

```C#
using Markdig;
using Markdig.Renderers.Normalize;

var text = @"- > p";

var pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().Build();
var markdownDocument = Markdown.Parse(text, pipeline);
var writer = new StringWriter();
var normalizer = new NormalizeRenderer(writer);

normalizer.Render(markdownDocument);

Console.WriteLine(writer.ToString());
```

Yields

```
- p
```

So I created this
- #953

Issue #320, opened 2019-03-12
NormalizeRenderer loses block quote formatting when used in lists


mircea-Hodis:
Hello,

I want to process a list of block quotes using the following code:

~~~c#
            var pipeline = new MarkdownPipelineBuilder().UseAdvancedExtensions().Build();
            var content =File.ReadAllText(sourcePath);

            var markdownDoc = Markdown.Parse(content, pipeline);

            using (var streamWriter = new StreamWriter(targetPath))
            {
                var renderer = new NormalizeRenderer(streamWriter);

                renderer.Render(markdownDoc);
            }
~~~

I'm using the following markdown as source:
~~~
List of blockquotes
1. > This is a quote in a list 1
2. > This is a quote in a list 2
3. > This is a quote in a list 3
~~~

The result is:
~~~
List of blockquotes

1. This is a quote in a list 1
2. This is a quote in a list 2
3. This is a quote in a list 3
~~~

I've expected the result to be the same but the block quote formatting is lost in the target file.
Is there any way to use Markdig and preserve both list and block quote formatting?

Regards.
Mircea

xoofx:
Yeah, `NormalizeRenderer` hasn't been developed to be fully supported/completed, it's a work in progress (e.g #155), so likely block quotes are not supported correctly. PR Welcome.

boxofyellow:
It looks like this problem may have later got reported in
- #482

I'm checking if this PR fixes it too 🤞 
- #953

boxofyellow:
Yes, it does look like #953 will fix this too.

PR #953, opened 2026-09-12
Fix indent on first list item (#482)

boxofyellow:
This PR fixes
- #482 (the issues I was originally looking at)
- #320 (the original report of the problem)

In short the `Markdig.Renderers.Normalize.ListRenderer` would include the text for the "list" part of the list item as text, and then push an `Indent` on the stack comprised on a string of spaces, and lastly then render all the children items.

This caused a problem when the first child block wanted to add its own `Indent` on the stack.  When that first item was a QuoteBlock, `previousWasLine` within `TextRendererBase` would still be false, causing its own indent to be "eaten" on the first line of the QuoteBlock.  As reported in the bug

```
- > p
```

Would get changed into

```
- p
```

At the same time if first child block was a sub list an extra new line would be inserted changing
```
- - a
```

Into
```
- 
  - a
```

To correct this problem, I used `Markdig.Renderers.Roundtrip.ListRenderer` as inspiration, and updated the Normalize version to move all of the "List" data into the `Indent`.  This did require creating a new constructor so that we could create `Indent`s that would have a specified "first line", and all subsequent uses of the `Indent` would yield a string the same width of the "first line", but made of spaces.  Additional a **_new public method was add_**, `PushHangingIndent` to allow pushing these new `Indent`s to the stack. 

This mostly addressed the problem.  However it created a problem for truly empty list items.  They had no children, and as a result `WriteChildren` would effectively be a no-op.  Again the Roundtrip version contains a solution, we detect this case and just write the empty string, which will respect the `Indent`.

There is one caveat with that, as noted in the two new tests `ListUnorderedEmpty` and `ListOrderedEmpty`.  With this implementation, the trailing space that follow the `OrderedDelimiter` or `BulletType` is always included, even if it's not in the input.  We could address this within our condition for the empty list, but I was not able to find a good way to detect the case when trailing space was in the input.  I assume it's better to add the space that will likely be required instead of eating it.  But that can easily be changed.

### DIFF
--- a/src/Markdig/Renderers/Normalize/ListRenderer.cs
+++ b/src/Markdig/Renderers/Normalize/ListRenderer.cs
@@ -38,12 +38,15 @@ public class ListRenderer : NormalizeObjectRenderer<ListBlock>
                 var item = listBlock[i];
                 var listItem = (ListItemBlock) item;
                 renderer.EnsureLine();
-
-                renderer.Write(index.ToString(CultureInfo.InvariantCulture));
-                renderer.Write(listBlock.OrderedDelimiter);
-                renderer.Write(' ');
-                renderer.PushIndent(new string(' ', IntLog10Fast(index) + 3));
-                renderer.WriteChildren(listItem);
+                renderer.PushHangingIndent($"{index.ToString(CultureInfo.InvariantCulture)}{listBlock.OrderedDelimiter} ");
+                if (listItem.Count == 0)
+                {
+                    renderer.Write(""); // trigger writing of indent
+                }
+                else
+                {
+                    renderer.WriteChildren(listItem);
+                }
                 renderer.PopIndent();
                 switch (listBlock.BulletType)
                 {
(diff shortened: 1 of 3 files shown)


src/Markdig/Renderers/TextRendererBase.cs
boxofyellow, line 192: This Null check looks like it might be a copy-paste error of
https://github.com/xoofx/markdig/pull/953/changes#diff-93312f2bb29ede6862c9f8cff9baf84ba584d58bd770d388a84381e3b523af22R183

And it should instead be checking to see if `lineSpecific` is null.  But I'm not sure since this is very much looking out for the case when `lineSpecific` is null
https://github.com/xoofx/markdig/pull/953/changes#diff-93312f2bb29ede6862c9f8cff9baf84ba584d58bd770d388a84381e3b523af22R109

src/Markdig/Renderers/Normalize/ListRenderer.cs
boxofyellow, line 29: Another difference I noticed we might want to drop this limited implication of ordered lists identifiers, and instead just use

https://github.com/xoofx/markdig/blob/b07b98e1b799db0daec908ebed91bf0cb62a00e1/src/Markdig/Renderers/Roundtrip/ListRenderer.cs#L31

